PR #5, opened 2015-03-11
Removed dependency on info.cukes:gherkin-jvm-deps and placed GenerateAst under test

tsundberg:
I am not sure if it was this way you intended to remove the dependency on info.cukes:gherkin-jvm-deps, but it is one way to do it.

While I was at it, I added some unit tests for GenerateAst and noticed that I think there are two bugs. There are two cases where a NullPointerException is thrown when the bad features are parsed. I am pretty certain that it should be some other exception. I am, however, not entirely sure where these other exceptions should be thrown. Look into the GenerateAstTest and you will see that NullPointerException is expected for two tests. This feels very wrong, but it is a documentation on the current behavior.

### DIFF
--- a/java/src/test/java/gherkin/GenerateAstTest.java
+++ b/java/src/test/java/gherkin/GenerateAstTest.java
@@ -1,0 +1,69 @@
+package gherkin;
+
+import org.junit.Test;
+
+import java.io.File;
+import java.io.FilenameFilter;
+import java.io.IOException;
+
+public class GenerateAstTest {
+    @Test
+    public void parse_all_good_features() throws IOException {
+        String[] args = getAllGoodTestFeatures();
+
+        GenerateAst.main(args);
+    }
+
+    @Test(expected = ParserException.CompositeParserException.class)
+    public void fail_on_inconsistent_cell_count() throws IOException {
+        String[] args = {"../testdata/bad/inconsistent_cell_count.feature"};
+
+        GenerateAst.main(args);
+    }
+
+    @Test(expected = ParserException.CompositeParserException.class)
+    public void fail_on_multiple_parser_errors() throws IOException {
+        String[] args = {"../testdata/bad/multiple_parser_errors.feature"};
+
+        GenerateAst.main(args);
+    }
+
+    @Test(expected = ParserException.CompositeParserException.class)
+    public void fail_on_single_parser_error() throws IOException {
+        String[] args = {"../testdata/bad/single_parser_error.feature"};
+
+        GenerateAst.main(args);
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void fail_on_invalid_language() throws IOException {
+        String[] args = {"../testdata/bad/invalid_language.feature"};
+
+        GenerateAst.main(args);
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void fail_on_unexpected_eof() throws IOException {
+        String[] args = {"../testdata/bad/unexpected_eof.feature"};
+
+        GenerateAst.main(args);
+    }
+
+    private String[] getAllGoodTestFeatures() throws IOException {
+        File dir = new File("../testdata/good/");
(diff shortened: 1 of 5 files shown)


aslakhellesoy:
Hi @tsundberg - great to see you getting involved with Gherkin 3!

I've been thinking a little more about removing the dependency on `info.cukes:gherkin-jvm-deps`, and I remember why I think we need to keep it.

The gson library is a commonly used library, so depending on it in gherkin is bound to cause version mismatch problems when userland code depends on a different version of gson.

We could build gherkin so that it shades gson, moving it to a different package, but this raises other problems:

During development of Cucumber-JVM it is often useful to mount Gherkin as a separate source tree in the IDE, making it possible to modify Gherkin while modifying Cucumber-JVM. In this case, gson would not be shaded, since we're compiling agains the non-shaded gson.

If Cucumber-JVM also needs gson (and it will), we won't be able to use the shaded package for gson, and that means we'll have to shade gson in the build of Cucumber-JVM too.

This leads to duplication of binaries and generally complicates things.

Therefore I think we need to keep the `info.cukes:gherkin-jvm-deps` package as it currently stands.

Regarding unit tests for the parser - that is certainly a nice improvement. We also need acceptance tests for this, using Make and diff - similar to how the acceptance tests work for *good* features.

What we would compare would be expected error messages. We might need to inject an special "test" error formatter so that the various implementations will print errors in the same way.

Can you update your PR to remove the gson changes? I'd like to merge in the unit tests.

tsundberg:
Hi!

I will update my pull request and only leave the first commit that places GenerateAst under test. You will have a new request before Tuesday.

/Thomas

java/src/test/java/gherkin/GenerateAstTest.java
```diff
@@ -0,0 +1,69 @@
...
+import java.io.File;
+import java.io.FilenameFilter;
+import java.io.IOException;
+
+public class GenerateAstTest {
+    @Test
+    public void parse_all_good_features() throws IOException {
+        String[] args = getAllGoodTestFeatures();
+
+        GenerateAst.main(args);
```
aslakhellesoy: I'm not a big fan of unit tests that print to STDOUT. I'd prefer to refactor `GenerateAst` to avoid this.

aslakhellesoy: I don't see a point in unit testing `GenerateAst` at all. It's already tested with the acceptance tests in `Makefile`. It would make more sense to have unit tests for `Parser`.

tsundberg:
I agree on the bad things about unit tests that writes to standard out. The tests as such doesn't do that, they are excersising GenerateAst and it prints to standard out.

I wasn't aware of the acceptance tests that verifies GenerateAst. I added some tests for it before I started to change which gson to use just to have a chance to know if I broke something or not.

I can't run the acceptance tests, running `make`fails for me with a unknown command `jq` on my mac book.

If it is better to not have any unit tests for GenerateAst and rely on the acceptance test, then let us just ignore this pull request. Or add a test that takes control over System.out.println and verifies that it gets called with the expected values.

tsundberg:
Should the error I found in `GenerateAst` be caught by `Parser`? 

I may take a look at that as well. In another pull request.

aslakhellesoy:
You need to install [jq](http://stedolan.github.io/jq/) as mentioned in `CONTRIBUTING.md`.

I think a couple of unit tests for the parser to verify that we get expected errors on bad input is enough. There will be more thorough cross-platform acceptance tests anyway.

aslakhellesoy:
> Should the error I found in GenerateAst be caught by Parser?

You mean the other way around right? No - let errors bubble through. GenerateAst is primarily used for testing, but I left it in the jar so people can run it on own code in case they find a bug.

tsundberg:
Yep, an error in the parser should trigger a Gherkin exception and bubble through. That was what I tried to mean and obviously wasn't clear about.

PR #7, opened 2015-03-31
add all errors to js CompositeParserException so it can be inspected later

chmontgomery:
useful when plugging into an ace editor worker


aslakhellesoy:
Thanks!

Issue #4, opened 2015-03-10
Finding the right way to provide the PHP implementation


stof:
This issue is meant to continue the discussion started in https://twitter.com/everzet/status/575285360403611648 in a better place than Twitter /cc @everzet @cyaranmcnulty

The current Gherkin parser available in PHP is https://github.com/Behat/Gherkin/. I see 2 different ways to provide the gherkin3-compatible parser for PHP:

1. keeping behat/gherkin as parser for PHP, but making it run the same acceptance tests than the cucumber/gherkin3 parsers (see #3 which was opened for this reason)
2. generate a PHP parser in this repo (using a different PHP namespace than ``Behat`` given that ``behat/gherkin`` is already at version 4 anyway and the generated library would probably not be a drop-in replacement)

The first solution means that all the maintenance of the PHP parser stays on the shoulders of the Behat team, but the acceptance tests would ensure consistency.
The following concerns the second solution

### Composer support

The PHP parser MUST be available as a [Composer](https://getcomposer.org/) package registered on [Packagist](https://packagist.org/) (otherwise everyone will hate the library, starting with the Behat team). However, Composer does not play well with repositories containing code for multiple languages:

- it expects to have the composer.json file at the root of the repository (not that annoying given that it is a single file, but still inconsistent with other languages)
- Composer will then download the Github archive (or clone the repo depending on the user preference) to get the PHP code, meaning PHP users will download lots of garbage code

A way to handle this is to have the generated PHP parser in a repo containing only the PHP library. There is 2 ways to achieve this:

- generating it in this repo and then maintaining a subtree repo, which will be used by Packagist. This requires some external server responsible for updating the subtree split automatically on each push to the main repo
- generating it only in the gherkin-php repo. This requires to think about regenerating this one when updating the parser definitions in the main repo, or having an automated update of the gherkin-php repo. It makes the maintenance work bigger when updating parser definitions here

### Impact on Behat

Currently, the ``Behat\Gherkin`` node classes (representing the AST) are exposed to userland code (especially the PyStringNode and the TableNode). So Behat itself would have 2 choices to use the new parser:

- break backward compatibility by dropping behat/gherkin entirely and making userland code use the AST objects generated by the new parser (and then hoping that the cucumber follows semver properly on the gherkin parser so that changes here don't break BC for Behat users by mistake).
[...]

PR #9, opened 2015-03-31
add build scripts to programmatically create files ace editor can consume

chmontgomery:
First, install the new tools by running `npm install`.

now, running `gulp` from `gherkin3/javascript` will generate a `dist/gherkin/` folder containing all the files which can be dropped wholesale into the [ace project under `lib/ace/mode/gherkin/`](https://github.com/chmontgomery/ace/tree/feature/gherkin-worker/lib/ace/mode/gherkin)

This change depends on https://github.com/cucumber/gherkin3/pull/7 and https://github.com/cucumber/gherkin3/issues/8

### DIFF
--- a/javascript/Makefile
+++ b/javascript/Makefile
@@ -55,6 +55,9 @@ dist/gherkin.js: lib/gherkin/parser.js ../LICENSE node_modules/.fetched
 	echo '*/' >> $@
 	./node_modules/.bin/browserify index.js --ignore-missing >> $@
 
+ace: lib/gherkin/parser.js ../LICENSE node_modules/.fetched
+	./node_modules/.bin/gulp
+
 dist/gherkin.min.js: dist/gherkin.js node_modules/.fetched
 	mkdir -p `dirname $@`
 	echo '/*' > $@
--- a/javascript/Makefile
+++ b/javascript/Makefile
@@ -13,7 +13,7 @@ all: .compared
 .compared: .built $(TOKENS) $(ASTS) $(ERRORS)
 	touch $@
 
-.built: lib/gherkin/parser.js lib/gherkin/dialects.json $(JAVASCRIPT_FILES) dist/gherkin.js dist/gherkin.min.js node_modules/.fetched
+.built: lib/gherkin/parser.js lib/gherkin/dialects.json $(JAVASCRIPT_FILES) dist/gherkin.js dist/gherkin.min.js dist/gherkin/parser.js node_modules/.fetched
 	./node_modules/.bin/mocha
 	touch $@
 
@@ -55,7 +55,7 @@ dist/gherkin.js: lib/gherkin/parser.js ../LICENSE node_modules/.fetched
 	echo '*/' >> $@
 	./node_modules/.bin/browserify index.js --ignore-missing >> $@
 
-ace: lib/gherkin/parser.js ../LICENSE node_modules/.fetched
+dist/gherkin/parser.js: lib/gherkin/parser.js ../LICENSE node_modules/.fetched
 	./node_modules/.bin/gulp
 
 dist/gherkin.min.js: dist/gherkin.js node_modules/.fetched
--- a/javascript/gulpfile.js
+++ b/javascript/gulpfile.js
@@ -1,0 +1,63 @@
+var gulp = require('gulp');
+var wrap = require('gulp-wrap-amd');
+var through = require('through2');
+var path = require('path');
+var fs = require('fs');
+
+function replaceAll(str, find, replace) {
+  return str.split(find).join(replace);
+}
+
+gulp.task('wrap:gherkin', function () {
+  gulp.src('./lib/**/*.js')
+    .pipe(through.obj(function (file, enc, cb) {
+      if (file.isNull()) {
+        cb(null, file);
+        return;
+      }
+
+      var fileName = path.basename(file.path);
+      var newFileContents = replaceAll(file.contents.toString(), "require('tea-error')", "require('./tea/error')");
+
+      if (fileName === 'token_matcher.js') {
+        // token_matcher.js requires a json file. To make the file browser compatible, we need to paste the
+        // contents of the json file into the js file
+        return fs.readFile(path.join(__dirname, 'lib/gherkin/dialects.json'), 'utf8', function (err, data) {
+          if (err) throw err;
+          file.contents = new Buffer(replaceAll(newFileContents, "require('./dialects.json')", data.trim()));
+          cb(null, file);
+        });
+      }
+
+      file.contents = new Buffer(newFileContents);
+      cb(null, file);
+    }))
+    .pipe(wrap({
+      exports: 'module.exports'
+    }))
+    .pipe(gulp.dest('./dist/'))
+});
+
+gulp.task('wrap:node_modules', function () {
+  gulp.src([
+    './node_modules/tea-error/lib/error.js',
+    './node_modules/tea-error/node_modules/tea-extend/lib/extend.js'
+  ])
+    .pipe(through.obj(function (file, enc, cb) {
+      if (file.isNull()) {
+        cb(null, file);
+        return;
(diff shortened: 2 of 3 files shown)


aslakhellesoy:
I'm a little skeptical about adding a whole new build system just to accommodate for a different project.

I'd prefer to do this without gulp. I just added a browserify build and added some notes to the JavaScript README. Do you think what I suggest there would work?

mattwynne:
I think that, in general, we need a way for things like editor plugins to be separate projects. Otherwise this project will grow into the same size of beast that Gherkin 2 was.

mattwynne:
Oh my bad, I've read the code now. Ignore me.

chmontgomery:
Initially I tried browserify but when putting the result of browserify through ace's [dryice](https://github.com/mozilla/dryice) build system it generates an invalid `worker-gherkin.js`. I'm not sure if this is a bug with dryice, with ace's implementation of it, or both. I've opened an issue for this problem under their repo: https://github.com/ajaxorg/ace/issues/2433

In the meantime, the only way I was able to get valid ace files created was taking your original src files and wrapping them with amd before dropping them into ace's src. In the end, browserify would be cleaner so we don't have to do any hackey string replacements. Unfortunately, as of now, I haven't found a way to do it.

As far as a different build system, I'm not tied to gulp. However, it would be trivial to tie the gulp build into your make file the same way you did browserify. It would be as simple as:

```
dist/gherkin.js: .compared node_modules/tea-error/lib-cov/error.js
	mkdir -p dist
	./node_modules/.bin/gulp
```

aslakhellesoy:
@chmontgomery can you check if 55d2419938f1aad5402753d145c8df6d7193266a works? The `javascript/dist/gherkin.js` file created by `make` should be sufficient.

chmontgomery:
@aslakhellesoy I just tried https://github.com/cucumber/gherkin3/commit/55d2419938f1aad5402753d145c8df6d7193266a and the ace build still generates invalid `worker-gherkin.js` files as described here: https://github.com/ajaxorg/ace/issues/2433. It seems to be a limitation of their build system... I also tried wrapping the file you generate with amd syntax in a couple of different ways with no success.

aslakhellesoy:
FYI - I have already wrapped the entry-point in AMD syntax: https://github.com/cucumber/gherkin3/blob/master/javascript/index.js

So the browserified `javascript/dist/gherkin.js` should already be AMD compatible - in theory. I haven't tested it. To narrow it down you could try to create a new example in `javascript/examples` that uses an AMD loader (require.js perhaps) to verify that it's indeed working correctly with AMD.

chmontgomery:
Most likely your generated file works in a stand alone amd example, the problem is with ace's build system. Dryice needs to be able to understand the file we're giving it so it can properly generate the different variations of the `worker-gherkin.js` file. I've tried every way I can think of to take your browserified file and pass it off to ace with no success. Unfortunately I don't have the time to work through ace's build file and/or dryice and provide a fix, hence this issue: https://github.com/ajaxorg/ace/issues/2433

aslakhellesoy:
They are not making it easy!

Out of curiosity - wouldn't it be easier to keep a Gherkin plugin for Ace outside the Ace project? Doesn't Ace provide a simple mechanism to plug in 3rd party plugins?

Codemirror makes that super easy. I've created a [3rd party Codemirror plugin](github.com/share/share-codemirror), and it lives its own life.

Shoehorning something into Ace sounds like an uphill struggle....

chmontgomery:
agreed, shoehorning ace is a losing battle... that's why I went with my original approach... taking your src files and wrapping them with a simple amd syntax. That seemed to play nice with ace.

I'm not an ace expert but they don't seem to have a plugin system. For other parsers they support like javascript, css, json, etc, they're checked into the ace repo directly under `lib/ace/mode/<language>`, just like your's would be.

The problem with supporting our own 3rd party library is we would still need to build ace compatible files. Ace generates build files for src, src-min, src-min-noconflict, and src-noconflict. Each of these versions are wrapped differently to work in various environments. We could copy ace's build system into a separate repo and modify it to our liking, but I'd prefer not to maintain something like that.

I've updated this PR to handle licensing properly so that the result, `dist/gherkin/**`, can be dropped whole-sale into the ace repo and run through their build system. I've added a new `make` target for this called `ace`

javascript/Makefile
```diff
@@ -55,6 +55,9 @@ dist/gherkin.js: lib/gherkin/parser.js ../LICENSE node_modules/.fetched
 	echo '*/' >> $@
 	./node_modules/.bin/browserify index.js --ignore-missing >> $@
 
+ace: lib/gherkin/parser.js ../LICENSE node_modules/.fetched
```
aslakhellesoy: The name of the target should not be `ace`, but the path to the file that gulp produces. Otherwise, `make` won't be able to detect that there is nothing to do.

chmontgomery: got it. Will update

chmontgomery:
updated js `Makefile` to only build when needed

aslakhellesoy:
I appreciate your effort to bring Gherkin to Ace, but your approach is introducing extra work for both the Gherkin team and the Ace team:

* The Gherkin team has to maintain build scripts only needed to accommodate for Ace
* The Ace team has to maintain copies of Gherkin in their own repo - ref comment by @nightwing in ajaxorg/ace#2429

I'll illustrate it with a dependency diagram:

```
    ┌────────┐  
    │        │  
    │        ▼  
┌───────┐  ┌───┐
│Gherkin│  │Ace│
└───────┘  └───┘
    ▲        │  
    │        │  
    └────────┘  
```

You have to remove the maintenance burden from the Gherkin and Ace teams by inverting the dependencies:

```
┌───────┐    ┌───────────┐    ┌───┐
│Gherkin│◀───│Gherkin-Ace│───▶│Ace│
└───────┘    └───────────┘    └───┘
```

You should create a separate GitHub repo where you put your `gherkin_worker.js` file, but you *don't* copy Gherkin source code into this repo. That's how JavaScript code was reused before package managers such as npm and bower become mainstream.

You'll have the freedom to use your favourite build tools, and you'll produce an Ace *plugin*. I did a quick search in the Ace docs for how to write plugins, and didn't find anything. That doesn't mean Ace doesn't have a plugin API - it just means you'll have to ask the Ace team for help.

When you distribute your plugin on npm and/or bower, I recommend you *don't* embed the Gherkin code there either, just declare a dependency in package.json/bower.json. Consumers of your plugin will then get Gherkin as a transitive dependency.

What do you think about this approach?

chmontgomery:
I think that makes a lot of sense. I originally wanted to do something similar but, as we mentioned earlier, fighting ace would be an uphill battle. I was just trying to fit into their system without too much friction.

We can close this PR and I'll pull the couple build scripts into my own repo where it can be published separately.

chmontgomery:
@aslakhellesoy can you publish gherkin3 to npm? right now all that's in npm is version 2.

aslakhellesoy:
Gherkin3 isn't stable yet, so publishing now would be premature. I don't expect significant changes, but I'm sure there will be some minor ones. We don't really know until we start replacing Gherkin2 with Gherkin3 in the various Cucumber implementations. That will start happening as soon as #12 is resolved.

Until Gherkin3 is released to npm you can depend on the git repo:
https://docs.npmjs.com/files/package.json#git-urls-as-dependencies

I'm glad we're making progress on this!

chmontgomery:
@aslakhellesoy I can't depend on this project from my package.json because this project's package.json is not at the top level of the project. Currently npm does not support depending on a package in a subdirectory of a repo: https://github.com/npm/npm/issues/2974

aslakhellesoy:
Oh I see. There is a similar problem for PHP as described in #4. We might need to split Gherkin up in several git repos.

aslakhellesoy:
@chmontgomery Now there is a standalone repo:

https://github.com/cucumber/gherkin-javascript

PR #20, opened 2015-04-07
Generic makefile

muhqu:
A more generalized version of the global `Makefile`.

I prefer pattern rules over explicit ones, but some do not. What do you think?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,49 +1,28 @@
-all: csharp/.compared go/.compared java/.compared javascript/.compared ruby/.compared
+
+MAKEFILES=$(wildcard */Makefile)
+
+all: $(patsubst %/Makefile,%/.compared,$(MAKEFILES))
 .PHONY: all
+%/.compared: %
+	cd $< && make
 
-clean:
-	cd csharp && make clean
-	cd go && make clean
-	cd java && make clean
-	cd javascript && make clean
-	cd ruby && make clean
+clean: $(patsubst %/Makefile,clean-%,$(MAKEFILES))
 .PHONY: clean
+clean-%: %
+	cd $< && make clean
 
-csharp/.compared:
-	cd csharp && make
-
-go/.compared:
-	cd go && make
-
-java/.compared:
-	cd java && make
-
-javascript/.compared:
-	cd javascript && make
-
-ruby/.compared:
-	cd ruby && make
-
-add-remotes:
-	git remote add gherkin-csharp     git@github.com:cucumber/gherkin-csharp.git
-	git remote add gherkin-go         git@github.com:cucumber/gherkin-go.git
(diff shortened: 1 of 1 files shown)


aslakhellesoy:
Great improvement! I merged and pushed, then realised that `push-subtrees` and `pull-subtrees` are broken:

```
$ make push-subtrees
git remote -v | grep -E 'gherkin-csharp\t' 2>/dev/null >/dev/null || \
	git remote add gherkin-csharp git@github.com:cucumber/gherkin-csharp.git
NAME=csharp
git subtree push --prefix=add-remote-csharp gherkin-add-remote-csharp master
'add-remote-csharp' does not exist; use 'git subtree add'
make: *** [push-subtree-csharp] Error 1
```

Can you push a fix for that @muhqu ?

muhqu:
already fixed with a6ae0ec16236ca634f01d2ae89593129a597a45d

muhqu:
@aslakhellesoy just pushed the go subtree https://github.com/cucumber/gherkin-go

muhqu:
@aslakhellesoy btw, the language specific travis-ci builds will not work, as the Makefiles in the language specific builds assume that the acceptance test-data is available in the upper directory (`../`).

It would be necessary to include the acceptance test data [and dialects.json and gherkin.berp] in the subtree repositories... Maybe the global Makefile should copy these into the language directories? …but then we have lot of duplicates. I guess the best way would be to have these files only copied for the subtree repository, but not the sub directory... If you know what I mean.

tooky:
On Tue, 7 Apr 2015, at 09:42 AM, Mathias Leppich wrote:
> @aslakhellesoy btw, the language specific travis-ci builds will not work,
> as the Makefiles in the language specific builds assume that the
> acceptance test-data is available in the upper directory (`../`).
> 
> It would be necessary to include the acceptance test data [and
> dialects.json and gherkin.berp] in the subtree repositories... Maybe the
> global Makefile should copy these into the language directories? …but
> then we have lot of duplicates. I guess the best way would be to have
> these files only copied for the subtree repository, but not the sub
> directory... If you know what I mean.

Sorry I've not completely kept up with the changes as its schoo
holidays, but it occurred to me that each of the implementations should
depend on 'something' that provides the acceptance tests? Rather than
the central repository depending on each of the implementations? I'm not
sure what that means for releases though?

Steve

aslakhellesoy:
Currently `make` will fail if you run it from a `gherkin-*` repo. I think that's fine. I'd expect contributors and maintainers to work in the `gherkin3` repo. Copying lots of files around (`gherkin.berp`, `testdata`...) seems like a lot of duplication. I'm already not super happy about copying the `dialects.json` and `LICENSE` around, but I suppose those are the bare minimum.

The `gherkin-*` repos are primarily there to accommodate for package managers who want code to live in a root directory.

WDYT?

PR #21, opened 2015-04-07
Switch to JSON AST in c#

gasparnagy:
Making a PR to test build before merge


aslakhellesoy:
I used to have interfaces and superclasses in the Java AST, then decided to ditch it. I think that in most cases, if a class hierarchy doesn't honour the [Liskov substitution principle](http://en.wikipedia.org/wiki/Liskov_substitution_principle), then inheritance is not needed.

I also found that by using GSON's reflective JSON generation, all I needed to control the JSON output was to name classes and fields. I expect it will be the same for Newton JSON.

What do you think about retrofitting the .NET AST to follow the current Java AST structure?

aslakhellesoy:
I'm contradicting myself wrt inheritance - there is a `Node` base class. The only reason I added it was to aid the JSON generation so all nodes have a `type` and `location`.

gasparnagy:
The .NET and the Java AST is almost the same already. The only difference is the casing of the property names. Such JSON formatting issues and also the `type` field can be done with contract resolvers in Newtonsoft JSON. This is implemented here: dotnet/Gherkin.AstGenerator/FeatureAstJSonContractResolver.cs. So I think generally we should be fine with the AST synchronization soon.

aslakhellesoy:
Ok. What do you think about removing the various interfaces in the Ast package? Are they really needed?

I tried to merge your branch, but I got a few merge conflicts - the master branch has been busy in the past 2 weeks!

Looking forward to having JSON support in .NET

gasparnagy:
i'll take care of the merging. I'll make a separate commit for the csharp->dotnet change.
i'll follow up on the interface question afterwards

gasparnagy:
finally I was able to rebase on master and merge...
@aslakhellesoy regarding the interfaces: once we will have the compiler, they will not be needed. I will remove them afterwards. I make a TODO entry for it.

PR #23, opened 2015-04-07
Cleanup dialects to use gherkin-languages.json everywhere

gasparnagy:
Let's see if build likes it...


gasparnagy:
@muhqu do you have any idea what is the problem with the go compilation here (and why was it working before?) this is on the `cleanup-dialects` branch, so you can also reproduce it.

muhqu:
@gasparnagy looks like there was some newline issue in the multiline `jq` script that was inline in the `Makefile`. I have now put the script into a separat file (with some more formatting), and now it's working. See 1abda632b46c75eaa36cba5a8c76ab33f407ef55.

gasparnagy:
@muhqu thx!

muhqu:
@gasparnagy no problem, sry for having checked in failing stuff in the first place… :-/

btw, shouldn't this cleanup also include the removal of `generate_dialects.js` ?

aslakhellesoy:
I'd like to keep it for now. There are a few unmerged PRs for new translations in Gherkin 2!

gasparnagy:
@muhqu What do i need to install (on Ubuntu) to be able compile the go version? I did `sudo apt-get install golang` but I get an error `./gherkin3.go:71: undefined: bufio.Scanner`. Could you maybe extend https://github.com/cucumber/gherkin3/blob/master/CONTRIBUTING.md?

muhqu:
In that case your GOROOT/GOPATH env vars seem to be misconfigured (or not set at all), as the compiler is not able to find a package from the stdlib. you might want to look [here](http://www.renyikhor.com/blog/installing-go-in-ubuntu-and-how-to-fix-gopath-goroot/).

gasparnagy:
@muhqu no luck. neither GOROOT/GOPATH was set. I've tried to set them (to what?) but it did not help... I don't really get the "working directory" concept though. Where do these env vars pointing for you?

muhqu:
@gasparnagy I'm on OSX and use the homebrew go. my GOROOT and GOPATH are not set, so they fallback to their defaults which are configured at compile-time in the homebrew recipe.

You can check with `go env` what is being used. If your go defaults GOROOT to a location that doesn't exist you might want to create a symlink on that location to point to the right location (where ever that is)... If you still have trouble with the apt-get version of go, you might just want to install the [binary distribution](https://golang.org/doc/install#tarball) wich defaults to `/usr/local/go`.

gasparnagy:
@muhqu OK. The apk-get version is go1, too old. I solved it by installing from binary.

muhqu:
@gasparnagy  :mortar_board:  :sparkles:

PR #29, opened 2015-04-08
Header and body in examples (WIP)

gasparnagy:
As discussed at commit https://github.com/cucumber/gherkin3/commit/e89c30e2ea612d5ac89e93b830c1df3c6a644b95 we switch back to separated Header/Body for examples.

I have created the expected AST files and the .NET code, but have no time now for the other languages, so if anyone want to pick it -> go ahead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: csharp
-mono: latest
+mono: 3.12.0
 env: TARGET_LANG=dotnet
 
 matrix:
--- a/dotnet/Gherkin/Ast/Examples.cs
+++ b/dotnet/Gherkin/Ast/Examples.cs
@@ -11,8 +11,8 @@ namespace Gherkin.Ast
         public string Keyword { get; private set; }
         public string Name { get; private set; }
         public string Description { get; private set; }
-        public TableRow Header { get; private set; }
-        public IEnumerable<TableRow> Body { get; private set; }
+        public TableRow TableHeader { get; private set; }
+        public IEnumerable<TableRow> TableBody { get; private set; }
 
         public Examples(Tag[] tags, Location location, string keyword, string name, string description, TableRow header, TableRow[] body)
         {
@@ -21,13 +21,13 @@ namespace Gherkin.Ast
             Keyword = keyword;
             Name = name;
             Description = description;
-            Header = header;
-            Body = body;
+            TableHeader = header;
+            TableBody = body;
         }
 
         IEnumerable<TableRow> IHasRows.Rows
         {
-            get { return new TableRow[] {Header}.Concat(Body); }
+            get { return new TableRow[] {TableHeader}.Concat(TableBody); }
         }
     }
 }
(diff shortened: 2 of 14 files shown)


gasparnagy:
@muhqu could you help on this with the golang?

muhqu:
Can look into it, but probably not before mid next week... Maybe @nightlyone wants to give it a try?

brasmusson:
OK, I have updated the remaining languages to use header and body (also the languages I never written any code in before; go and javascript).

gasparnagy:
@brasmusson fantastic. this project is a great fun for language explorers...

aslakhellesoy:
Awesome @brasmusson

muhqu:
@gasparnagy & @brasmusson :+1: 

@aslakhellesoy I just read through the comments at [`e89c30e`](https://github.com/cucumber/gherkin3/commit/e89c30e2ea612d5ac89e93b830c1df3c6a644b95#commitcomment-10608443) and wanted to reply to your last one...

>  I'm convinced. Feel free to change it to header+body. (I prefer body over rows, but not fussed)

I'm absolutely for the separation of header and body rows, which makes much sense for the *Examples*, but I think the AST elements should better named *TableHeader* + *TableBody*. It's just I think it doesn't feel natural to each over a property solely called "*Body*" to retrieve *Row(s)*... I would expect a property called "*Body*" to refer to a potentially rather large string or byte-type of data. Naming it *TableBody* would make it more obvious what kind of data we're dealing with here...

aslakhellesoy:
I agree @muhqu - feel free to change it

mattwynne:
> I'm absolutely for the separation of header and body rows, which makes much sense for the Examples, but I think the AST elements should better named TableHeader + TableBody. It's just I think it doesn't feel natural to each over a property solely called "Body" to retrieve Row(s)... I would expect a property called "Body" to refer to a potentially rather large string or byte-type of data. Naming it TableBody would make it more obvious what kind of data we're dealing with here…
> 
I agree.

For a DataTable (multiline step argument) can we use TableBody too, or should that be a different type? Headers don’t make sense for those.

muhqu:
@mattwynne sure *TableHeader* doesn't make sense for the *TableData* Step argument. But I don't think that having *TableBody* instead of *Rows* as a property is an improvement.

This is what we have for *TableData* (step argument) now...
``` js
{ type: "Step", ... , argument: {
  type: "DataTable", rows: [ { type: "TableRow", .. }, { type: "TableRow", .. }, ... ]
}}
```

…and this is what I think you're proposing.
``` js
{ type: "Step", ... , argument: {
  type: "DataTable", tableBody: [ { type: "TableRow", .. }, { type: "TableRow", .. }, ... ]
}}
```
 …is that correct @mattwynne?

mattwynne:
Ignore me. I think I was getting confused between types and properties.

muhqu:
@mattwynne :hear_no_evil:

muhqu:
@gasparnagy ur also fine with it as it is?

I think it's ready to get merged and pushed to the subtrees... no?

aslakhellesoy:
LGTM. Do it!

muhqu:
affirmative

muhqu:
`make push-subtrees` :white_check_mark:

gasparnagy:
@muhqu @brasmusson thx!

PR #30, opened 2015-04-08
go: experimental AST schema visualization

muhqu:
**no need to merge** …just wanted to share what nice things you can do with go and it's nice stdlib. 

`ast-dot-diagram.go` parses [ast.go](https://github.com/cucumber/gherkin3/blob/master/go/ast.go) and outputs the a [graphviz](http://www.graphviz.org/) digraph which can then be rendered with `dot`...

screenshot

/cc @aslakhellesoy @gasparnagy

### DIFF
--- a/go/ast-dot-diagram/ast-dot-diagram.go
+++ b/go/ast-dot-diagram/ast-dot-diagram.go
@@ -1,0 +1,88 @@
+package main
+
+import (
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io/ioutil"
+	"os"
+	"strings"
+)
+
+func main() {
+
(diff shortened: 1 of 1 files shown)


gasparnagy:
nice

aslakhellesoy:
Have you seen the [AST diagram in the README](https://github.com/cucumber/gherkin3/blob/master/README.md#ast)?

I'm not sure I see what additional value this brings - apart from being auto-generated. The layout isn't better, and the diamonds in the composition lines are at the wrong end ;-)

aslakhellesoy:
@muhqu now that you're here - I'm usually in the #cucumber IRC room on freenode if you want to chat...

PR #33, opened 2015-04-24
Rebuild generators on change

nightlyone:
by telling make that it needs them for doing their acceptance tests

This now exercises the acceptance tests again by simply typing `make` after changing a Go file.

@muhqu Please take a look


muhqu:
sure makes sense :clap:

PR #34, opened 2015-04-27
travis.yml: try out some stuff...

muhqu:


aslakhellesoy:
I think we can safely close this one now...

PR #35, opened 2015-04-29
Gracefully report parser error when file not gherkin

tooky:
I've started taking a look at using gherkin3 with cucumber-ruby-core,  and I noticed that we weren't handling empty Gherkin documents or trying to parse non-gherkin documents gracefully. This PR adds some acceptance tests for these errors, and includes a fix for ruby/javascript.


aslakhellesoy:
Wouldn't it be simpler to have getTags() & friends check if the argument is null/undefined?

tooky:
@aslakhellesoy yes - except we then call methods on the arguments returned. I had a quick stab at creating a null-object node, but it wasn't right.

PR #38, opened 2015-05-12
Markdown support

aslakhellesoy:
I've wanted Cucumber to support Markdown as an alternative syntax to Gherkin for a while. I think it could be beneficial in several areas:

* Ability to embed images alongside scenarios (UI mockups, various sketches and diagrams)
* Markdown has a lot of mindshare

This PR currently has a start of an implementation. It's surprisingly easy to add thanks to the modular architecture - just a new `TokenMatcher` implementation!

There aren't any automated tests yet, but there is a manual one:

```
mvn compile && bin/gherkin-generate-ast ../testdata/good/minimal.md | jq "."
```

Some thoughts about reporting results from Cucumber if we go forward with this. Each markdown documents would get rendered to a HTML file in a reports directory. Results would be written as a JavaScript file, appended into the same HTML file at the end. This JavaScript would modify the DOM: 

* Results: add css classes for status (success/failed/pending/undefined/skipped)
* Stack traces: add dom nodes when a step fails
* Attachments/embedded data: add `` tags

In order to do this, the HTML nodes must have information about line number. This can be added when the Markdown is rendered (before scenarios run). We'd use an existing Markdown processor and post-process the generated HTML by scanning both the markdown source and the generated HTML - should be fairly easy.

I'm excited about this!

### DIFF
--- a/java/gherkin-java.razor
+++ b/java/gherkin-java.razor
@@ -53,10 +53,6 @@ public class @Model.ParserClassName<T> {
         @foreach(var rule in Model.RuleSet.TokenRules)
         {<text>        @rule.Name.Replace("#", ""),
 </text>}        ;
-
-        public static TokenType cast(RuleType ruleType) {
-            return TokenType.values()[ruleType.ordinal()];
-        }
     }
 
     public enum RuleType {
@@ -89,37 +85,35 @@ public class @Model.ParserClassName<T> {
         }
     }
 
-    public T parse(String source) {
-        return parse(new StringReader(source));
+    public T parse(String source, ITokenMatcher tokenMatcher) {
+        return parse(new StringReader(source), tokenMatcher);
     }
 
-    public T parse(Reader source) {
+    public T parse(Reader source, ITokenMatcher tokenMatcher) {
         AstBuilder<T> builder = new AstBuilder();
-        return parse(source, builder);
+        return parse(source, builder, tokenMatcher);
     }
 
-    public T parse(Reader source, IAstBuilder<T> builder) {
+    public T parse(Reader source, IAstBuilder<T> builder, ITokenMatcher tokenMatcher) {
         ParserContext context = new ParserContext(
                 new TokenScanner(source),
-                new TokenMatcher(),
+                tokenMatcher,
                 builder,
                 new LinkedList<Token>(),
                 new ArrayList<ParserException>()
         );
 
-        startRule(context, RuleType.@Model.RuleSet.StartRule.Name);
+        startRule(context, RuleType.Feature);
         int state = 0;
         Token token;
-        do
-        {
+        do {
             token = readToken(context);
             state = matchToken(state, token, context);
-        } while(!token.isEOF());
+        } while (!token.isEOF());
 
-        endRule(context, RuleType.@Model.RuleSet.StartRule.Name);
+        endRule(context, RuleType.Feature);
 
-        if (context.errors.size() > 0)
-        {
+        if (context.errors.size() > 0) {
             throw new ParserException.CompositeParserException(context.errors);
         }
 
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -56,6 +56,12 @@
             <version>1.10.19</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.github.rjeschke</groupId>
+            <artifactId>txtmark</artifactId>
+            <version>0.13</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/java/src/main/java/gherkin/GenerateAst.java
+++ b/java/src/main/java/gherkin/GenerateAst.java
@@ -14,8 +14,9 @@ public class GenerateAst {
 
         for (String fileName : args) {
             InputStreamReader in = new InputStreamReader(new FileInputStream(fileName), "UTF-8");
+            Parser.ITokenMatcher tokenMatcher = TokenMatcherFactory.getTokenMatcher(fileName);
             try {
-                Feature feature = parser.parse(in);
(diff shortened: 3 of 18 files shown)


mattwynne:
Can you explain, with an example, how you'd express a scenario in markdown? It's not clear to me from reading the code.

aslakhellesoy:
@mattwynne Oops here you go: 30d5f6ff30a54fe6b2123dd6c9b601e6601aac3c

muhqu:
@aslakhellesoy I think it would make a lot more sense to use an existing Markdown parser to get an AST of the Markdown. And then just implement a 'pickles' interface for that AST.

Markdown has so many variations... 
``` markdown
# headline

## headline

 * item 1
 * item 2

headline
========

headline
--------

 - item 1
 - item 2
```

muhqu:
…or consider multiline list items...
``` markdown
*   This is a list item with two paragraphs.

    This is the second paragraph in the list item. You're
only required to indent the first line. Lorem ipsum dolor
sit amet, consectetuer adipiscing elit.

*   Another item in the same list.
```

aslakhellesoy:
@muhqu I'm not sure every language has good markdown parsers that produce ASTs. I suspect many of them will just produce HTML text. We should make design decisions that are easy to port to a new language.

I think it would be easy to support different header styles (event the underlined ones), and we don't need to support multiline bullet lists. We can also say we only support the `##` style headers.

mattwynne:
I have a more general comment, now I've seen the syntax.

I've thought about this quite a bit in the past, and what I'd really like to see is a way to mark up any kind of regular prose as executable.

For example:

```markdown
# Shouting to multiple listeners

When multiple listeners are involved, all of them will receive the shout.

For example: Suppose Sean is stood 50 metres from Lucy, and another user named Larry is 10 metres away from Sean. So overall they're spaced like this:

<link to image>

When Sean shouts "Hello everyone", then we'd expect that both Lucy and Larry will hear Sean's message.
```

We could use links to mark the parts of the prose that are executable steps. Perhaps something like this:

```text
For example: Suppose [Sean is stood 50 metres from Lucy](cucumber://step), and another [user named Larry is 10 metres away from Sean](cucumber://step). So overall they're spaced like this:
```

muhqu:
@mattwynne great example. I like the idea of the using markdown-*markup* to highlight what actually is the test data. :clap: 

:bulb: You can even go a step further and just use a regular markdown-html processor and then use a CSS selector like `a[href="cucumber://step"]` to extract all the *steps*... a post-processor would then be able to add class names like `passed, skipped, failed, ignored` to the `<a>` elements.

aslakhellesoy:
That's interesting @mattwynne! Would this mean that the format we parse should really be the HTML? I mean, people could use Asciidoc if they want - we just parse the resulting HTML?

Have you thought about what would be scenario delimiters? Headers?

Any thoughts about DataTables and DocStrings? Can you give an example?

muhqu:
@aslakhellesoy scenario delemiters, could be easily found by
 1. find all `<a href="cucumber://step">`
 1. for each of them bubbling up the DOM until you find a sibling `<h1>` or `<h[1-5]>` element.
 1. group steps (`<a>`) by their matched headline element...

muhqu:
i'm sure some nodejs wizkid would be able to do that in a single line of jQuery ;-)

mattwynne:
I would be hesitant to describe it as HTML format, since HTML is not a nice syntax for writing in, but yeah, you could think about it that way.

http://concordion.org/ was one of the things that inspired this line of thinking.

What I've heard from Concordion users is that the way their glue is implemented (inline attributes in HTML) is a maintenance nightmare, but people love the readability of the docs.

mattwynne:
To your other points Aslak, yes I think a header (containing executable steps) could be used as the scenario delimiter. I haven't thought about DocStrings or DataTables at all.

tooky:
How about Markdown with something like the Maruku meta-data syntax? http://maruku.rubyforge.org/proposal.html

aslakhellesoy:
I'm not suggesting that users *write* HTML, but I still think it might make sense that the parser *reads* HTML. This gives us a lot of flexibility - the HTML can be generated with Markdown, Asciidoc, Wysiwyg editors, whatevs!

I'm starting to think basing the parser on Markdown (as I've done in this PR) is not ideal.

Parsing the HTML serverside is easy - there are tons of good libs out there with excellent css selector/xpath support.

So: the good parts of concordion and none of the bad parts.

Thoughts on that?

mattwynne:
Steve, how would you see Maruku working for steps in arbitrary prose exactly? Could you give an example?

tooky:
I think the idea of using HTML as the actual parsed language is interesting - there are many different ways of producing HTML so it's easy to support multiple markup object.

I think links will work great for steps - it makes sense as it becomes possible for the report API to have clickable links showing the code behind the steps.

I'm not sure about using H2 to mean TestCase though, and we will need a way to link tables / text as DataTable / DocString. Maybe something like the maruku extensions could add this metadata?

```markdown
# Using markdown to specify behaviour

Markdown allows for far more natural language to be used when
specifying the behaviour of a system. Encouraging teams to actually
think about the language without constraining them with `Given`,
`When`, `Then`.

## Examples

### DocStrings are linked using metadata {: data-cucumber=scenario}

[The following text will be used as a DocString for this step](cucumber://step):

  > This is some multiline text of arbitary length, following the markdown format.
  > We will need to decide how we specify line breaks - following HTML or preformatted
  > conventions... but that could be based on the type of block element used?
  {: data-cucumber=docstring }
```

dg-ratiodata:
BTW: You're not limited to `maruko` if you're going to use ruby. `kramdown` supports [this](http://kramdown.gettalong.org/syntax.html#attribute-list-definitions) as well. You may want to abstract that markdown thing away and use some abstraction library like [tilt](https://github.com/rtomayko/tilt) instead.

PR #42, opened 2015-06-18
Added getters to pickles.

SquirrelGrip:
Added getters to the pickle classes in order to access the final variables.

### DIFF
--- a/java/src/main/java/pickles/Pickle.java
+++ b/java/src/main/java/pickles/Pickle.java
@@ -33,4 +33,8 @@ public class Pickle {
     public List<PickleLocation> getSource() {
         return source;
     }
+
+    public List<PickleTag> getTags() {
+        return tags;
+    }
 }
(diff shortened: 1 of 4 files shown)


aslakhellesoy:
Thanks for your contribution!

Please beware that this part of the API is experimental and subject to change.

With that in mind, please provide the following:

* why do you need this?
* tests
* implementation/tests in the other languages

Cheers,
Aslak

SquirrelGrip:
Hi Aslak,
I have only added getters to existing Java Pickle classes in order to access the private final member variables in the Pickle classes.

Tests…How do you test a getter? You usually don’t if they are just returning a value. They don’t contain any logic. None of the existing tests failed.

All other languages have exposure to underlying variables and private scope is not an issue. The change is a trivial change but helps in using the gherkin parser in java.

I am using the Gherkin 3 to parse feature file and find definitionSteps defined in a set of glue classes. The scenario(pickle) is then executed in it own thread. 

Adrian

aslakhellesoy:
> in order to access the private final member variables in the Pickle classes.

Sorry - I should have been more clear. I understand it's to access the fields - what I'd like to understand is what you need the fields for, and now I know :-)

Since this API is experimental and very likely to change or even be removed I'd prefer not to merge this in.

PR #44, opened 2015-07-02
Ruby: One cell to little when no newline after example table

brasmusson:
If there is not blank or newline after the last cell of the last row in the example table, then the last parsed table row will have one cell to little, when using the ruby version of Gherkin3. This in turn causes a inconsistent cell count exception.

Several of the feature files for Cucumber contain this type of example tables from using the following constructs in the feature files:
```
 Given a file named "features/f.feature" with:
    """
    <snip>
      Examples:
      | Value           |
      | Also Irrelevant |
    """
```

### DIFF
--- a/ruby/lib/gherkin3/gherkin_line.rb
+++ b/ruby/lib/gherkin3/gherkin_line.rb
@@ -36,7 +36,11 @@ module Gherkin3
     def table_cells
       column = @indent + 1
       items = @trimmed_line_text.split('|')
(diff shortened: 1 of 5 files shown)


brasmusson:
As it turned out to be a really small change, that made the Ruby implementation behave as the other language implementations, I went ahead and merged this PR.

aslakhellesoy:
So the other implementations were already doing this correctly? Nice!

PR #45, opened 2015-07-02
Update JavaScript Example

danielpacak:
Please correct me if I'm wrong, but this is the current API, where `gherkinDoc` is source as string.


aslakhellesoy:
This was updated independently, but thanks anyway!

PR #48, opened 2015-07-07
fixed errors - added Bosnian

10-io:


brasmusson:
@10-io Just a tip, you do not have to create a new PR when you want to add commit or do other changes. A PR is connected to its base branch, so adding, removing, squashing commits to the branch or rebasing is will be reflected in the PR.

PR #50, opened 2015-07-09
Escaped DocString content

mattwynne:
See https://github.com/cucumber/cucumber-ruby/issues/863#issuecomment-119888052 and https://github.com/cucumber/cucumber-ruby-core/pull/93

I've added what I think is the right failing test, but I'm still hunting around for the right way to implement the solution. Any tips?

### DIFF
--- a/dotnet/Gherkin/TokenMatcher.cs
+++ b/dotnet/Gherkin/TokenMatcher.cs
@@ -52,7 +52,7 @@ namespace Gherkin
         public bool Match_Other(Token token)
         {
             var text = token.Line.GetLineText(indentToRemove); //take the entire line, except removing DocString indents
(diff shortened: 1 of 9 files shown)


brasmusson:
@mattwynne I would look [here in the token_matcher.rb](https://github.com/cucumber/gherkin3/blob/master/ruby/lib/gherkin3/token_matcher.rb#L72-94) where the normal doc strings tokens are matched.

mattwynne:
Thanks for the clue @brasmusson I think I've nailed that. @gasparnagy @aslakhellesoy can you review before I start trying this on the other languages?

mattwynne:
OK I think this is ready to merge.

aslakhellesoy:
LGTM. Do it!

brasmusson:
Done!

PR #65, opened 2015-08-08
Reset ast builder and token matcher state (fix for #62)

### LINKED ISSUE #62: Python: Cannot parse multiple times
```python
from gherkin3.token_scanner import TokenScanner
from gherkin3.parser import Parser

parser = Parser()
print parser.parse(TokenScanner("Feature: 1"))
print parser.parse(TokenScanner("Feature: 2"))
```

Results in:

```python
{'language': 'en', 'keyword': u'Feature', 'tags': [], 'comments': [], 'location': {'column': 1, 'line': 1}, 'scenarioDefinitions': [], 'type': 'Feature', 'name': u'1'}
{'language': 'en', 'keyword': u'Feature', 'tags': [], 'comments': [], 'location': {'column': 1, 'line': 1}, 'scenarioDefinitions': [], 'type': 'Feature', 'name': u'1'}
```

Note the name hasn't changed from 1 to 2.  Am I missing something in how this is supposed to be used?

aslakhellesoy:
A similar change to Python (code and unit test) needs to be done in all languages:

* [x] Python
* [x] Java (I made some other unrelated changes as well)
* [x] Go
* [x] Ruby
* [x] .NET
* [x] JavaScript

fixed #62

### DIFF
--- a/dotnet/Gherkin.Specs/ParserErrorsTest.cs
+++ b/dotnet/Gherkin.Specs/ParserErrorsTest.cs
@@ -1,12 +1,9 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 using Gherkin.Ast;
 using Gherkin.AstGenerator;
+using Newtonsoft.Json;
 using NUnit.Framework;
 
 namespace Gherkin.Specs
@@ -34,7 +31,79 @@ namespace Gherkin.Specs
                 var expectedErrorsText = LineEndingHelper.NormalizeLineEndings(File.ReadAllText(expectedErrorsFile));
                 Assert.AreEqual(expectedErrorsText, errorsText);
             }
+        }
 
+        [Test]
+        public void TestFeatureAfterParseError()
+        {
+            var tokenMatcher = new TokenMatcher();
+            var parser = new Parser(new AstBuilder<Feature>());
+            var jsonSerializerSettings = new JsonSerializerSettings();
+            jsonSerializerSettings.Formatting = Formatting.Indented;
+            jsonSerializerSettings.NullValueHandling = NullValueHandling.Ignore;
+
+            try
+            {
+                parser.Parse(new TokenScanner(new StringReader(@"# a comment
+Feature: Foo
+  Scenario: Bar
+    Given x
+      ```
+      unclosed docstring")), tokenMatcher);
+                Assert.Fail("ParserException expected");
+            }
+            catch (ParserException)
+            {
+            }
+            var parsingResult2 = parser.Parse(new TokenScanner(new StringReader(@"Feature: Foo
+  Scenario: Bar
+    Given x
+      """"""
+      closed docstring
+      """"""")), tokenMatcher);
+            var astText2 = LineEndingHelper.NormalizeLineEndings(JsonConvert.SerializeObject(parsingResult2, jsonSerializerSettings));
+
+            string expected2 = LineEndingHelper.NormalizeLineEndings(@"{
+  ""Tags"": [],
+  ""Location"": {
+    ""Line"": 1,
+    ""Column"": 1
+  },
+  ""Language"": ""en"",
+  ""Keyword"": ""Feature"",
+  ""Name"": ""Foo"",
+  ""ScenarioDefinitions"": [
+    {
+      ""Tags"": [],
+      ""Location"": {
+        ""Line"": 2,
+        ""Column"": 3
+      },
+      ""Keyword"": ""Scenario"",
+      ""Name"": ""Bar"",
+      ""Steps"": [
+        {
+          ""Location"": {
+            ""Line"": 3,
+            ""Column"": 5
+          },
+          ""Keyword"": ""Given "",
+          ""Text"": ""x"",
+          ""Argument"": {
+            ""Location"": {
+              ""Line"": 4,
+              ""Column"": 7
+            },
+            ""ContentType"": """",
+            ""Content"": ""closed docstring""
+          }
+        }
+      ]
+    }
+  ],
+  ""Comments"": []
+}");
+            Assert.AreEqual(expected2, astText2);
         }
     }
 }
--- a/dotnet/Gherkin.Specs/SuccessfulParsingTests.cs
+++ b/dotnet/Gherkin.Specs/SuccessfulParsingTests.cs
@@ -26,16 +26,15 @@ namespace Gherkin.Specs
         public void TestMultipleFeatures()
         {
             var tokenMatcher = new TokenMatcher();
-            var astBuilder = new AstBuilder<Feature>();
-            var parser = new Parser<Feature>();
+            var parser = new Parser(new AstBuilder<Feature>());
             var jsonSerializerSettings = new JsonSerializerSettings();
             jsonSerializerSettings.Formatting = Formatting.Indented;
             jsonSerializerSettings.NullValueHandling = NullValueHandling.Ignore;
 
-            var parsingResult1 = parser.Parse(new TokenScanner(new StringReader("Feature: Test")), tokenMatcher, astBuilder);
-            var astText1 = JsonConvert.SerializeObject(parsingResult1, jsonSerializerSettings);
-            var parsingResult2 = parser.Parse(new TokenScanner(new StringReader("Feature: Test2")), tokenMatcher, astBuilder);
-            var astText2 = JsonConvert.SerializeObject(parsingResult2, jsonSerializerSettings);
+            var parsingResult1 = parser.Parse(new TokenScanner(new StringReader("Feature: Test")), tokenMatcher);
+            var astText1 = LineEndingHelper.NormalizeLineEndings(JsonConvert.SerializeObject(parsingResult1, jsonSerializerSettings));
+            var parsingResult2 = parser.Parse(new TokenScanner(new StringReader("Feature: Test2")), tokenMatcher);
+            var astText2 = LineEndingHelper.NormalizeLineEndings(JsonConvert.SerializeObject(parsingResult2, jsonSerializerSettings));
 
 	    string expected1 = LineEndingHelper.NormalizeLineEndings(@"{
   ""Tags"": [],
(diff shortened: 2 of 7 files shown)


dirkrombauts:
I can have a look at this one too ... what's it about?

aslakhellesoy:
This PR is a fix for #62, caused by a bug in `AstBuilder`

brasmusson:
The root cause for #62 is that the AstBuilder can carry state over from the parsing of one feature to the next. The TokenMatcher has the same problem, but it is harder to disclose - a parse error with an open doc sting separator is required. I have failing tests and fixed for Python and Ruby - but I need to reconcile the changes with the changes here so far.

aslakhellesoy:
@brasmusson would it make sense to DI the TokenMatcher as well and reset its state at the end of a parse? Maybe even do it on a parse/lex error?

brasmusson:
@aslakhellesoy The exiting of the parse method can be by a successful parse, or (if `stop_at_first_error is true`) by an raising an exception in any of the `match_token_at_<nr>`, or (if `stop_at_first_error is false`) by raising an exception in `parse`, for that reason resetting the state of AstBuilder and TokenMatcher at the start of the parsing feels safer with respect to catching all cases reset is needed before parsing the next feature.

brasmusson:
Now all language implementations are updated so no state can leak between different parsings in the TokenMatcher or AstBuilder. After looking at and working on the different language implementations, there are some questions to consider:
* Should the TokenMatcher be passed to the Parser, or should it as in the [Java implementation](https://github.com/cucumber/gherkin3/blob/30b926d949358626245005edda57e797a4c4e9b3/java/src/main/java/gherkin/Parser.java#L114) be create fresh in the parse method each call? Creating it fresh avoids any state to be leaked in it between parsings.
* Should the AstBuilder (and if applicable the TokenMatcher) be passed into the parser at construction, or into the `parse` method as an argument? Now most implementations pass them into the parser at construction (following the example of @aslakhellesoy with the AstBuilder in the Java and Python implementation). I took the path of minimal change with the languages I am less familiar with (Go and C#) and kept the AstBuilder and TokenMatcher as arguments to the `parse` method (it felt more important to get to "green" first, and then refactor - if needed).

aslakhellesoy:
@brasmusson good question regarding the TokenMatcher (whether it should be passed to the constructor or the parse method).

I just pushed a change to Java where it's passed to the constructor, but I'm not 100% sure it's a good idea, and that it might be better to pass it in via parse().

Ultimately, what I want to be able to do is to make it possible to use Markdown and/or HTML (with special `data-gherkin` attributes) as alternative syntaxes to Gherkin. For example:

```markdown
<!-- bullets are steps -->
## This is a scenario
* Given I have 42 cukes in my belly

<!-- bold is steps -->
## This is a scenario
I invite you to consider a context where *I have 42 cukes in my belly* (let's assume small ones)
```

Or HTML:

```html
<div data-gherkin=scenario>
I invite you to consider a context where <span data-gherkin=step>I have 42 cukes in my belly</span> (let's assume small ones) 
</div>
```

The idea is to author executable specifications in a richer format using an editor such as [ProseMirror](http://prosemirror.net/), [woofmark](http://bevacqua.github.io/woofmark/) or [Quill](http://quilljs.com/). This would allow embedding of illustrations, and generally richer text, while some elements of the text can be marked up to be scenarios and steps.

So: How can we structure Gherkin3 so it's relatively easy to plug in a new scanner/lexer and reuse everything else (parsing, error checking, AST building).

I'm thinking we'd have a `MarkdownTokenMatcher` or even a `HTMLTokenMatcher`. In that case, it probably makes more sense to pass the TokenMatcher to the parse method.

Am I making any sense?

mattwynne:
@aslakhellesoy interesting that you're thinking of dealing with those new formats inside the Gherkin library. I think @tooky and I were envisaging that we'd keep Gherkin for Feature/Scenario/Given/When/Then etc, and write new lexer / parser *libraries* (peers of Gherkin) that plug into the the core / pickles.

mattwynne:
I'm not sure it makes sense to build a Gherkin AST from a markdown doc - it might limit things artificially. Better to just build the more abstract tree of test cases / test steps I think.

aslakhellesoy:
I'm not saying Markdown/HTML support necessarily needs to be in the Gherkin library itself - they could be separate libraries that can plug into Gherkin3. In order for that to work, Gherkin3 needs to define an API where alternative scanners can be plugged in. My current thinking is that the `TokenScanner` and `TokenMatcher` are candidates.

Of course, there are many possible layers where we could plug this in. The lowest level is the lexer level (what I'm exploring here). We could also plug in at the AST builder level (generate an AST straight from markdown without going via Gherkin3's parser/AstBuilder). And finally we could bypass Gherkin altogether and produce Pickles/TestCase (what the Gherkin3 compiler will produce form AST).

I'm interested in plugging in at the lowest possible level in order to reuse some of the complicated code. Compiling backgrounds and Scenario Outlines is a little complicated, so I'd like to reuse the compiler.

Detecting grammatical errors (for example, a tag sitting between steps) is the job of the parser, and that would be nice to reuse too (as well as the building of the AST the compiler needs).

That brings us to the scanner level. I'd like to build a MarkdownScanner (in a separate lib) that I can just pass to Gherkin3 like so:

```ruby
parser = Parser.new
featureAst = parser.parse(markdown, MarkdownLexer.new)
```

mattwynne:
Yeah I see, so what we're talking here is more a dialect of Gherkin - a different medium for expressing what are still Gherkin scenarios. Makes sense then.

aslakhellesoy:
Exactly - in a way you can think of it as different representations of the same thing.

dirkrombauts:
Is there still anything to be done on the dotnet side? Or does commit  30b926d fix everything?

brasmusson:
@dirkrombauts What I think still need to be done in the dotnet implementation, is to move the builder argument of the Parser to ctor argument so that the [setup for the added test cases](https://github.com/cucumber/gherkin3/blob/30b926d949358626245005edda57e797a4c4e9b3/dotnet/Gherkin.Specs/SuccessfulParsingTests.cs#L28-L38) can become.
```
    ...
    var parser = new Parser<Feature>(new AstBuilder<Feature>());
    ...
    var parsingResult1 = parser.Parse(new TokenScanner(new StringReader("Feature: Test")), tokenMatcher);
    ...
```
The conclusion both form the discussion above and an IM chat I had with @aslakhellesoy is that the matcher argument shall be passed to the `parse` method (so different matchers can be used for different file types in the a file list parsed).

Maybe the added test cases do not fit perfectly in the SuccessfullParsingTest fixture, which previously focused on checking that the good acceptance test features parsed without error, and is better moved to a new fixture.

The [makefile](https://github.com/cucumber/gherkin3/blob/reset-ast-builder-state/dotnet/Makefile#L18) executes the test cases in the Gherkin.dll, but it contains not test cases. Wouldn't it be better to use the Gherkin.Specs.dll which contains test cases?

dirkrombauts:
@brasmusson I should have some time tomorrow afternoon to take a look at the constructor argument. And I'll change the makefile too.

dirkrombauts:
@brasmusson do those two unit tests you added pass on your machine? I'm talking about TestMultipleFeatures and TestFeatureAfterParserError.

On my machine they are failing because of newline issues - "\n" is expected but the result (on my machine) is "\r\n". I could change the expected result but I expect the tests would then fail on linux and mac.

Shall I put in placeholders and replace them with System.Environment.Newline?

dirkrombauts:
@brasmusson forget about the last comment. I simply needed to call `LineEndingHelper.NormalizeLineEndings` on the string that is asserted.

dirkrombauts:
And I sent a pull request (#67)

brasmusson:
@dirkrombauts You did what I said. What I also thought (only in my own mind an obvious consequence of what I said) was that the generated [`Parser<T>`](https://github.com/cucumber/gherkin3/blob/00f1a9188cb307d7d1b0a14e5d092b8223c0cb59/dotnet/Gherkin/Parser.cs#L76) should have an `IAstBuilder` as ctor argument. This would also mean that the `astBuilder` is moved from the `ParserContext` and stored directly in the `Parser<T>`. Then also the [`GenerateTokens`](https://github.com/cucumber/gherkin3/blob/00f1a9188cb307d7d1b0a14e5d092b8223c0cb59/dotnet/Gherkin.TokensGenerator/TokensGenerator.cs#L11-L12) would create a parser like:
```
    var parser = new Parser<object>(new TokenFormatterBuilder());
```

dirkrombauts:
@brasmusson yes I thought that `Parser<T>` should be adapted too. Question is: how do I adapt the generated code thing?

brasmusson:
@dirkrombauts [gherkin-csharp.razor](https://github.com/cucumber/gherkin3/blob/00f1a9188cb307d7d1b0a14e5d092b8223c0cb59/dotnet/gherkin-csharp.razor) is the input file for the generation, and the [makefile](https://github.com/cucumber/gherkin3/blob/00f1a9188cb307d7d1b0a14e5d092b8223c0cb59/dotnet/Makefile#L46-L47) contains the commands to perform the generation. In this case the ares needed to be changed are at the [class definition and parse method](https://github.com/cucumber/gherkin3/blob/00f1a9188cb307d7d1b0a14e5d092b8223c0cb59/dotnet/gherkin-csharp.razor#L63-L93) and where the [tokenMatcher is used](https://github.com/cucumber/gherkin3/blob/00f1a9188cb307d7d1b0a14e5d092b8223c0cb59/dotnet/gherkin-csharp.razor#L182).

dirkrombauts:
@brasmusson OK looks doable later today

dirkrombauts:
@brasmusson I updated my pull request.

dirkrombauts:
@brasmusson as you will see I updated the class definition and parse methods, but I don't think there's anything to be done in the lines where the astBuilder is used: the lines you linked to all use the ParsingContext, and the ParsingContext is set in the Parse method.

brasmusson:
@dirkrombauts The reason I pointed you in that direction is to be consistent with the change in Java implementation in (the first commit in this PR) 033be13. There, when the `astBuilder` is passed to the ctor, it is also removed from the `parserContext` and instead accessed directly from the `Parser` field (and it is no longer possible to pass an `astBuilder` to the parse method). An argument to remove the `astBuilder` from the `parserContext` could be to focus the intent of the `parserContext` on holding the items that can change during the execution of the `parse` method, or than can change between invocations of the `parse` method, whereas the `astBuilder` is constant during the lifetime of the `Parser` instance.

dirkrombauts:
@brasmusson is the Java version stable and can I use it for "inspiration" for the C# version?

brasmusson:
@dirkrombauts I think that the Java (and the other) implementation have become quite stable (the specific commit I referred to is not the whole Java change - but it does nearly all change with respect to the AstBuilder). On the other hand if you fix @aslakhellesoy [comment](https://github.com/cucumber/gherkin3/pull/67#discussion_r36921241), I think you pretty much end up consistent with the Java implementation.

PR #67, opened 2015-08-11
Reset ast builder state (dotnet)

dirkrombauts:
I found some time to do the changes today :-)

I added the necessary constructors and methods to `Gherkin/Parser.Extensions.cs`. Is that the right place? If they should go into the generated code (`Parser<T>`): how should I do that?


dotnet/Gherkin/Parser.cs
```diff
@@ -89,7 +101,12 @@ protected class ParserContext
 
         public T Parse(ITokenScanner tokenScanner)
         {
-            return Parse(tokenScanner, new TokenMatcher(), new AstBuilder<T>());
+            return Parse(tokenScanner, new TokenMatcher(), this.AstBuilder);
+        }
+
+        public T Parse(TokenScanner tokenScanner, TokenMatcher tokenMatcher)
```
aslakhellesoy: Should be `ITokenMatcher` I believe.

dotnet/Gherkin/Parser.cs
```diff
@@ -89,7 +101,12 @@ protected class ParserContext
 
         public T Parse(ITokenScanner tokenScanner)
         {
-            return Parse(tokenScanner, new TokenMatcher(), new AstBuilder<T>());
+            return Parse(tokenScanner, new TokenMatcher(), this.AstBuilder);
+        }
+
+        public T Parse(TokenScanner tokenScanner, TokenMatcher tokenMatcher)
+        {
+            return Parse(tokenScanner, tokenMatcher, this.AstBuilder);
```
aslakhellesoy: Don't pass `this.AstBuilder` around. It's an instance variable. Also, remove it from the next method and from `ParserContext`.

dotnet/Gherkin/Parser.cs
```diff
@@ -75,8 +75,20 @@ public enum RuleType
...
+        }
+
+        public Parser(AstBuilder<T> astBuilder)
+        {
+            this.AstBuilder = astBuilder;
+        }
+
         public bool StopAtFirstError { get; set;}
 
+        public AstBuilder<T> AstBuilder { get; }
```
brasmusson: Isn't `readonly` the keyword in C# to say that this instance variable cannot change/is constant?

dotnet/Gherkin/Parser.cs
dirkrombauts: Passing instance variables is my usual strategy when dealing with overloads. I'm happy to learn about alternative strategies.

In this concrete case, though: if you mean to remove the overload that takes an AstBuilder by "remove it from the next method" then of course the problem of having to deal with the overload disappears.

dotnet/Gherkin/Parser.cs
dirkrombauts: That depends on the version of C# :-) C# 6 (introduced with Visual Studio 2015) allows me to write a readonly property like that. If you feel that this would cause problems for the build servers or many other contributors than I will change it to a `readonly` field.

dotnet/Gherkin/Parser.cs
dirkrombauts: Yes, thank you.

dotnet/Gherkin/Parser.cs
brasmusson: I think expressing/clarifying intent is important. By making something a constant at the declaration says clearly "this does not change", so when the language support saying that - then why not use it? I do not know C# idioms that well to say that it is better to change it to a `readonly` field.

dirkrombauts:
```
public string Property { get; }
```

is (as far as I know) equivalent to

```
private readonly string myField;
public string Property { get { return myField; } }
```

But will probably change my version to the version with a `readonly` field because I just realized that I have now idea how closely mono follows the development of C#.

brasmusson: Given that the `astBuilder` is only used inside the `Parser` class, I wonder if it really needs a `get`, it seems to me that the `astBuilder` field could be declared like the [`dialectProvider` field in the TokenMatcher](https://github.com/cucumber/gherkin3/blob/master/dotnet/Gherkin/TokenMatcher.cs#L11).

dirkrombauts:
@aslakhellesoy @brasmusson I think I addressed all your concerns. Could you have another look please?

aslakhellesoy:
Great work @dirkrombauts !

dirkrombauts:
:-)

PR #70, opened 2015-08-16
Updated Docs for version 3.1.0

thetutlage:


aslakhellesoy:
Thanks for the update, but I prefer to make those arguments optional. I've updated the docs slightly differently.

PR #74, opened 2015-09-01
(Python) Close file descriptors

Zearin:
Also:

* Factor out `DIALECT_FILE_PATH` from `DIALECTS`
* Move `DIALECTS` from a class variable to a module variable

AFTER making these changes, I discovered a nearly line-for-line mirror of them in the Ruby implementation.  Knowing this, I named the `DIALECT_FILE_PATH` variable after its Ruby implementation equivalent (originally the variable was named something else).

So…happy accident. :-)

### DIFF
--- a/python/gherkin3/dialect.py
+++ b/python/gherkin3/dialect.py
@@ -1,14 +1,22 @@
+import io
 import os
 import json
 
 
+DIALECT_FILE_PATH = os.path.join(
+    os.path.dirname(__file__),
+    'gherkin-languages.json'
+    )
+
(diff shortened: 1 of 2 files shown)


brasmusson:
Thanks. Since calling `line.decode('utf-8')` when using Python 2 is not needed when using `io.open`, then the boolean flag `python2` is no longer needed in token_scanner.py, so I [removed it](https://github.com/cucumber/gherkin3/commit/886b6ed3b5df8a3fc8c0370647b7438098e2c919).

Zearin:
@brasmusson I meant to do that, but I guess I forgot.  Thanks!

PR #85, opened 2015-09-05
Add docstring comments for AstNode and AstBuilder

Zearin:
* Docstring content inspired by text in README.md
* Each implementation has the appropriate docstring (sometimes with a minimal amount of their preferred flavor of markup)

### DIFF
--- a/dotnet/Gherkin/AstBuilder.cs
+++ b/dotnet/Gherkin/AstBuilder.cs
@@ -5,6 +5,14 @@ using Gherkin.Ast;
 
 namespace Gherkin
 {
+    /// <summary>
+    /// The AstBuilder (used by the Parser) builds the AST (Abstract Syntax Tree).
+    /// The AST is comprised of AstNode objects (or “nodes”).
+    /// 
+    /// The implementation is simple objects without behaviour, only data. 
+    /// It's up to the implementation to decide whether to use classes or just basic collections, 
+    /// but the AST must have a JSON representation (this is used for testing).
+    /// </summary>
     public class AstBuilder<T> : IAstBuilder<T>
     {
         private readonly Stack<AstNode> stack = new Stack<AstNode>();
--- a/dotnet/Gherkin/AstNode.cs
+++ b/dotnet/Gherkin/AstNode.cs
@@ -4,6 +4,19 @@ using System.Linq;
 
 namespace Gherkin
 {
+    /// <summary>
+    /// Nodes are created by the AstBuilder, and are the building blocks of the AST (Abstract Syntax Tree). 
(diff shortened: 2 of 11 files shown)


Zearin:
Oops!  I know what’s breaking the build.  Just a moment and I’ll fix it up…

Zearin:
w00t!  How bout it, @aslakhellesoy ?

Zearin:
ping

Zearin:
@aslakhellesoy ping

aslakhellesoy:
I'm starting to think that these comments are going to make it difficult to maintain the several implementations of this library. The footprint of change becomes quite big when comments are duplicated like this. It's already hard enough to keep code in sync.

I think I'd prefer a link to a central place.

WDYT?

Zearin:
> I'm starting to think that these comments are going to make it difficult to maintain the several implementations of this library. The footprint of change becomes quite big when comments are duplicated like this. It's already hard enough to keep code in sync.
> 
> I think I'd prefer a link to a central place.
> 
> WDYT?

I share your concern.  (Hell, I share it on _principle_.)  I’d prefer a central place, too.  

On the other hand, my intent was to provide a short “synopsis” bit of documentation, consisting of text that was unlikely to change.  I was trying to limit my documentation PRs to just the main logical components.  I figured that as long as I kept them to a short summary, they would remain accurate even if a few details of the implementation changed.  

…

But yes—I nevertheless agree that a central place would be best.  I can’t think of anything off the top of my head, though.  

Maybe something clever involving git hooks?

aslakhellesoy:
How about just linking to a place on GitHub? For example a new `API.md` file:

```markdown
## AstBuilder

The AstBuilder (used by the Parser) builds the AST (Abstract Syntax Tree).
The AST is comprised of AstNode objects (or "nodes").
 
The implementation is simple objects without behaviour, only data. 
It's up to the implementation to decide whether to use classes or just basic collections, 
but the AST must have a JSON representation (this is used for testing).
```

Then in code:

```javascript
/**
 * See https://github.com/cucumber/gherkin3/blob/master/API.md#AstBuilder
 */
```

WDYT?

Zearin:
It’s better than nothing…but I’d gladly put in the extra effort if there were a way to include the docs alongside the source code.  Not exhaustive, document-every-single-thing documentation…just the summaries of the main parts.  

I’ve worked on projects that kept all their documentation on a website before, and it slowly eroded my enthusiasm for contributions.  

I can’t offer anything better off the top of my head, but this has _got_ to have happened enough times to enough people that someone has figured out a way to do exactly what we want…

aslakhellesoy:
Thanks for this. I prefer not to merge this since it makes maintenance harder (too much to sync)

PR #87, opened 2015-09-09
(Python) Various minor improvements

Zearin:
Quick summary: 

Moved things that aren’t essential to a class from class-level to module level; use modern `str.format()` over concatenation; reorder methods to read the `TokenMatcher` class a bit easier; and don’t use mutable default arguments.  :)


Zearin:
Crap…I’m not sure what’s breaking the build.  Gimme a little while…I’ll fix this!

Zearin:
I’m having some irritating unicode errors that don’t make sense to me. Reopening with a new Pull Request containing the non-str.format changes.

PR #88, opened 2015-09-10
(Python) Minor refactoring (and 1 potential issue fix)

Zearin:
Small stuff; see the full commit messages for details.

The change that avoids using `[]` as a default argument, however, is a (potentially) big deal.  Its commit message contains a link to an explanation, far more well-written than I could muster. :)

### DIFF
--- a/python/gherkin3/token_matcher.py
+++ b/python/gherkin3/token_matcher.py
@@ -35,7 +35,9 @@ class TokenMatcher(object):
         return False
     
     def _set_token_matched(self, token, matched_type, text=None,
-                           keyword=None, indent=None, items=[]):
+                           keyword=None, indent=None, items=None):
+        if items is None:
+            items = []
         token.matched_type = matched_type
         token.matched_text = text.rstrip('\r\n') if text is not None else None  # text == '' should not result in None
         token.matched_keyword = keyword
--- a/python/gherkin3/token_matcher.py
+++ b/python/gherkin3/token_matcher.py
@@ -50,6 +50,9 @@ class TokenMatcher(object):
     def _unescaped_docstring(self, text):
         return text.replace('\\"\\"\\"', '"""')
     
+    #
+    #   TITLE LINE matches
+    #
     def match_FeatureLine(self, token):
         return self._match_title_line(token, 'FeatureLine', self.dialect.feature_keywords)
 
@@ -65,7 +68,10 @@ class TokenMatcher(object):
 
     def match_ExamplesLine(self, token):
         return self._match_title_line(token, 'ExamplesLine', self.dialect.examples_keywords)
-
+    
+    #
+    #   ALL OTHER matches
+    #
     def match_TableRow(self, token):
         if not token.line.startswith('|'):
             return False
--- a/python/gherkin3/token_matcher.py
+++ b/python/gherkin3/token_matcher.py
@@ -3,10 +3,13 @@ from .dialect import Dialect
 from .errors import NoSuchLanguageException
 
 
-class TokenMatcher(object):
-    LANGUAGE_RE = re.compile(r"^\s*#\s*language\s*:\s*([a-zA-Z\-_]+)\s*$")
+DEFAULT_DIALECT_NAME = 'en'
+LANGUAGE_RE = re.compile(r"^\s*#\s*language\s*:\s*([a-zA-Z\-_]+)\s*$")
 
-    def __init__(self, dialect_name='en'):
+
+class TokenMatcher(object):
+
+    def __init__(self, dialect_name=DEFAULT_DIALECT_NAME):
         self._default_dialect_name = dialect_name
         self._change_dialect(dialect_name)
         self.reset()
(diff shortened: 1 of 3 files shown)


Zearin:
@aslakhellesoy ping

aslakhellesoy:
> @aslakhellesoy ping

I appreciate your contributions, but please be more patient. I do get notifications without the extra ping. I think a ping is more appropriate if nothing has happened in a week or two.

Zearin:
> I appreciate your contributions, but please be more patient. I do get notifications without the extra ping. I think a ping is more appropriate if nothing has happened in a week or two.

My apologies.  No ill intended.

brasmusson:
When I wrote the Python implementation I ensured that `pep8 --max-line-length=99` did not give any errors and warnings. That was intended to give at least some level of cleanness and consistency. The generated parser.py was made an exception, where even longer lines were accepted (to not complicate the razor file), and in a couple of places I added end line comment which made the lines longer than 99 characters (it seemed the least bad thing to do at the time). This PR introduce a set of warnings from `pep8 --max-line-length=99`, almost exclusively about trailing whitespace

python/gherkin3/token_matcher.py
```diff
@@ -17,6 +20,41 @@ def reset(self):
         self._indent_to_remove = 0
         self._active_doc_string_separator = None
 
+    def _change_dialect(self, dialect_name, location=None):
```
brasmusson: There are different strategies for ordering functions and methods. Public methods before private methods is one - it was basically the one used in this class previously - even though `_match_DocStringSeparator` was an exception to that rule. "In call graph order" is another strategy (like `_match_DocStringSeparator` follows `match_DocStringSeparator` from which is called).
Here I do not see the strategy used. It is not "methods should be declared before they are used", because `_change_dialect` is placed after `reset` which calls it.

The main objection I have is the move of `_unescaped_docstring` it is only used in `match_Other`, so I think it should be kept closer to that method.

python/gherkin3/token_matcher.py
```diff
@@ -50,13 +91,11 @@ def match_StepLine(self, token):
             title = token.line.get_rest_trimmed(len(keyword))
             self._set_token_matched(token, 'StepLine', title, keyword)
             return True
-
         return False
 
     def match_Comment(self, token):
         if not token.line.startswith('#'):
             return False
-
```
brasmusson: No, it not necessary to have a blank line separating the guard clause from the main work of the method, but do you think it is so bad, that it motivates going through the whole method and remove them? I try hard not to ask because you propose changes in code I originally wrote, but from the perspective I try to have when evolving existing code - "be consistent with what is already there" (unless there is a very good reason).

python/gherkin3/token_matcher.py
aslakhellesoy: I agree with @brasmusson - I don't see how the new proposed ordering improves readability or consistency with the other implementations.

python/gherkin3/token_matcher.py
aslakhellesoy: This whole PR seems to be change for the sake of change. Is there anything substantial I'm missing? Like a bugfix?

brasmusson: In theory the change from using `[]` as default argument in https://github.com/Zearin/gherkin3/commit/176cae0e23f6071a994a477a15d1972b35d1f9ff could fix a problem. In practice I do not think it can happen, since `token.matched_items` is never modified, only iterated over. Still, to avoid warning bells ringing when people reads that code, it could be wise to use the default value `None` instead. When I look at that method now I would change the assignment to: `token.matched_items = items or []` in order to make `token.matched_items` possible to iterate over at all times - instead of using an if-statement that conditionally assigns `[]` to `items`. That keeps `_set_token_matched` focused on assigning the token attributes.

Using `str.join` instead of string concatenation with `+` as in https://github.com/Zearin/gherkin3/commit/0b1b7c55c85fbf73e463fecb6b0a590c15128e76 is more efficient (https://waymoot.org/home/python_string/). On the other hand since `TokenFormatterBuilder` is only used in the acceptance tests, maybe efficiency is not the most important thing there.

Zearin:
> This PR introduce a set of warnings from pep8 --max-line-length=99, almost exclusively about trailing whitespaces, Please fix those warnings. 

Yeah, I am awful about trailing whitespace. :(  Sorry!  Will fix shortly.

python/gherkin3/token_matcher.py
Zearin: > No, it not necessary to have a blank line separating the guard clause from the main work of the method, but do you think it is so bad, that it motivates going through the whole method and remove them? 

No, it doesn’t.  I did not intend any negative statement about those spaces.  Sorry.  :(

> I try hard not to ask because you propose changes in code I originally wrote, but from the perspective I try to have when evolving existing code - "be consistent with what is already there" (unless there is a very good reason).

I understand this mindset.  I will do my best to respect it from now on.  

I see that many of my changes are not to your liking, or @aslakhellesoy ’s.  I love this project, and I’d like to contribute.  But it’s frustrating to discover that my PRs seem violate a lot of undocumented expectations of the code.  

Have I wasted everyone’s time with this PR?  I don’t know.  Maybe…maybe not.  But I think it would be a good idea to document things like your PEP8 `--max-line-length`, and the “be consistent with what is already there (unless there is a very good reason)” mentality to review.  There is nothing wrong with that mentality, but there _are_ projects who are perfectly okay with the sorts of changes in this PR. Others share the mentality, but have different ideas about what is a “good reason”.  Where is a contributor supposed to learn where you stand on this?

Zearin:
Gah.  After reading the comments above, I think this PR displeases the core team more than it pleases it.  Perhaps it’s better not to merge this.

I will aim for smaller changes in the future.

PR #95, opened 2015-09-16
Fix Ruby example in Readme

vincent-psarga:
A scanner is required when parsing a feature file.
parser_spec use the same process: https://github.com/cucumber/gherkin3/blob/master/ruby/spec/gherkin3/parser_spec.rb

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,10 @@ var feature = parser.Parse("Feature: ...");
 ```ruby
 # Ruby
 require 'gherkin3/parser'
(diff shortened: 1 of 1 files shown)


aslakhellesoy:
Actually, it's the code that needs to be fixed, not the docs. It should works like this:

https://github.com/cucumber/gherkin3/blob/master/javascript/lib/gherkin/parser.js#L51-L54
https://github.com/cucumber/gherkin3/blob/master/python/gherkin3/parser.py#L60-L61

PR #97, opened 2015-09-23
Fix readme for ruby and fix missing require statement

dg-ratiodata:
I found some small issues with gherkin3. 

1. The example does not work for ruby - at least not in a C&P-fashion. I changed this. Now ruby users can just start an irb and paste the code there.

2. Furthermore `StringIO` is used, but needs to be loaded first. I added the require statement directly to the class-file of the tokenscanner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,10 @@ var feature = parser.Parse("Feature: ...");
 ```ruby
 # Ruby
 require 'gherkin3/parser'
+require 'gherkin3/token_scanner'
+
 parser = Gherkin3::Parser.new
-feature = parser.parse("Feature: ...")
+feature = parser.parse(Gherkin3::TokenScanner.new("Feature: ..."))
 ```
 
 ```javascript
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -44,8 +44,10 @@ var feature = parser.Parse("Feature: ...");
 ```ruby
 # Ruby
 require 'gherkin3/parser'
+require 'gherkin3/token_scanner'
+
 parser = Gherkin3::Parser.new
-feature = parser.parse("Feature: ...")
+feature = parser.parse(Gherkin3::TokenScanner.new("Feature: ..."))
```
aslakhellesoy: To make the API more intuitive to use, we want `parse` to be able to take a scanner as well as a `String` or `IO`. See #95. Could you fix the code instead of the docs please? It should work the way the docs say.

dg-ratiodata: Done. I sent you another PR. Now `TokenScanner` is able to extract io from another `TokenScanner`-instance to make it backward compatible.

PR #98, opened 2015-09-23
Make parser take strings as well

dg-ratiodata:
Follow up to #97

* `Parser#parse` accepts strings, io, file
* `TokenScanner#new` gets io from another `TokenScanner`-instance to make it backward compatible
* `Parser#parse` uses `TokenScanner` directly to give different inputs - io, file, string, token_scanner - a sane API

### DIFF
--- a/ruby/lib/gherkin3/parser.rb
+++ b/ruby/lib/gherkin3/parser.rb
@@ -1,6 +1,7 @@
 # This file is generated. Do not edit! Edit gherkin-ruby.razor instead.
 require_relative 'ast_builder'
 require_relative 'token_matcher'
+require_relative 'token_scanner'
 require_relative 'errors'
 
 module Gherkin3
@@ -63,7 +64,9 @@ module Gherkin3
       @ast_builder = ast_builder
     end
 
-    def parse(token_scanner, token_matcher=TokenMatcher.new)
+    def parse(input, token_matcher=TokenMatcher.new)
+      token_scanner = input.is_a?(TokenScanner) ? input : TokenScanner.new(input)
+
       @ast_builder.reset
       token_matcher.reset
       context = ParserContext.new(
--- a/ruby/lib/gherkin3/token_scanner.rb
+++ b/ruby/lib/gherkin3/token_scanner.rb
@@ -1,8 +1,9 @@
-require_relative 'token'
-require_relative 'gherkin_line'
+require 'gherkin3/scanners/file_scanner'
+require 'gherkin3/scanners/string_scanner'
+require 'gherkin3/scanners/io_scanner'
+require 'gherkin3/scanners/fail_scanner'
 
 module Gherkin3
-  
   # The scanner reads a gherkin doc (typically read from a .feature file) and 
   # creates a token for line. The tokens are passed to the parser, which outputs 
   # an AST (Abstract Syntax Tree).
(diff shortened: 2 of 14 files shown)


brasmusson:
`parser.rb` is generated from [`gherkin-ruby.razor`](https://github.com/cucumber/gherkin3/blob/master/ruby/gherkin-ruby.razor#L60), so you need to change it and then let the Makefile re-generate `parser.rb`.

ruby/lib/gherkin3/parser.rb
```diff
@@ -63,9 +63,12 @@ def initialize(ast_builder=AstBuilder.new)
       @ast_builder = ast_builder
     end
 
-    def parse(token_scanner, token_matcher=TokenMatcher.new)
+    def parse(input, token_matcher=TokenMatcher.new)
```
aslakhellesoy: This is inconsistent with the other implementations. parse(scanner) should work too.

ruby/lib/gherkin3/token_scanner.rb
```diff
@@ -11,15 +13,23 @@ module Gherkin3
...
     def initialize(source_or_path_or_io)
       @line_number = 0
+
       if String === source_or_path_or_io
         if File.file?(source_or_path_or_io)
           @io = File.open(source_or_path_or_io, 'r:BOM|UTF-8')
         else
           @io = StringIO.new(source_or_path_or_io)
         end
+      elsif TokenScanner === source_or_path_or_io
```
aslakhellesoy: Shouldn't be allowed to pass a scanner to a scanner's ctor. See how it's done in js/python/java

ruby/lib/gherkin3/parser.rb
dg-ratiodata: It does! https://github.com/dg-ratiodata/gherkin3/commit/d9085a51cc39f516795f2ef2b0b5f6bdcb415a81#diff-a5a70ea60d3411ec934a6b3412ff8af6R31

dg-ratiodata:
I'm adding some more specs. So please wait a bit before merging.

ruby/lib/gherkin3/token_scanner.rb
```diff
@@ -11,15 +13,23 @@ module Gherkin3
   # to look for Gherkin keywords for the associated language. The keywords are defined 
   # in gherkin-languages.json.
   class TokenScanner
+    protected
+
+    attr_reader :io
```
aslakhellesoy: Please remove this. Not needed when `Parser.parse` looks at the type of the first arg to decide whether it's already a scanner or if it needs to wrap a scanner around a string or io.

dg-ratiodata: Ok. I can do this, thought that should be better part of the token scanner and not any business of the parser.

You already have a "if"-statement here https://github.com/dg-ratiodata/gherkin3/commit/d9085a51cc39f516795f2ef2b0b5f6bdcb415a81#diff-a5a70ea60d3411ec934a6b3412ff8af6L17 I see no good reason why to introduce another one in the parser.

dg-ratiodata: Additionally:
The scanner raises an error to make the user aware about the content it accepts.

dg-ratiodata:
@brasmusson How do I run the make-command only for ruby? I've got no mono/dotnet installed. Is there any documentation for this?

ruby/lib/gherkin3/token_scanner.rb
aslakhellesoy: The scanner is able to scan a String or IO. Nothing else (especially not another scanner).

The parser is able to get tokens from a scanner, and for the sake of API friendliness, it can create a scanner out of a string.

Maybe the whole if statement should move to the parse method. I also don't like that a String can mean file path.

dg-ratiodata:
@brasmusson Ok. Found a way.

aslakhellesoy:
> How do I run the make-command only for ruby? 

```
cd ruby
make
```

> I've got no mono/dotnet installed. Is there any documentation for this?

Just download and install mono.
http://www.mono-project.com/download/

ruby/lib/gherkin3/token_scanner.rb
dg-ratiodata: > I also don't like that a String can mean file path.

I'm not sure if this breaks "your" API, but why not say a path needs to be a `Pathname`?

> The scanner is able to scan a String or IO. Nothing else (especially not another scanner).
> Maybe the whole if statement should move to the parse method. 

Why not using polymorphism for that?

* StringScanner
* IOScanner
* FileScanner

I will provide an implementation for that, then you can choose if you like it. :-)

dg-ratiodata:
@aslakhellesoy What do you think about the current implementation?

aslakhellesoy:
On its own I think it's nice. However, consistency is more important - see 74965391b4457d4a9b5358270aee4ccd9c3d76f0.

I'm afraid this code is too inconsistent with the other implementations.

maxmeyer:
@aslakhellesoy If you read this evening: You can close this one. I opened 2 new PRs with my private account.

PR #102, opened 2015-09-29
Python package: add trove classifiers

koterpillar:
These are used by https://caniusepython3.com/project/gherkin3 and other PyPI-related tools.


aslakhellesoy:
Out of curiosity - what does this enable people to do?

koterpillar:
Besides the site, you also get shown on "Python 3 Packages" on https://pypi.python.org/pypi/, so people know Gherkin isn't holding up their migration.

PR #107, opened 2015-10-10
Swift

concertman:
@aslakhellesoy here is the beginnings of the Swift parser. It still needs work before it will compile and parse.


aslakhellesoy:
Thanks! I merged it to a new [swift](https://github.com/cucumber/gherkin3/tree/swift) branch for now. Hopefully more people can come and help us move this forward.

PR #110, opened 2015-10-15
Objective-C version

LiohAu:
After the Swift comes the Objective-C version based on the donet version :)

Tokens generation seems to be ok, but AST generation tests fails on the "descriptions.feature" case (because of the "description" property..), it's probably a trim call or a special character replacement that is incorrect...


aslakhellesoy:
I'm impressed! You seem to have done this in just a couple of days which is a new speed record so far I think.

I have some questions/comments that I hope you can answer. Please keep in mind that I have zero experience with Objective C, Cocoa and XCode. (I'm on a Mac and I have XCode installed).

## Build artefacts

When I run `make`, my build artefacts don't end up under `./build` (as the `bin/*` scripts assume), but under `/Users/aslakhellesoy/Library/Developer/Xcode/DerivedData/Gherkin-bnumctuldkbcrbbwtxzbuivmrjzu/Build/Products/Debug/TokensGenerator` and `/Users/aslakhellesoy/Library/Developer/Xcode/DerivedData/Gherkin-bnumctuldkbcrbbwtxzbuivmrjzu/Build/Products/Debug/AstGenerator`.

Can you make a change to ensure they end up in the right place?

## What OS is this for?

I was hoping this would run on both iOS and OS X, but I see a lot of OSX mentioned. Can this code be built for iOS at all?

## Using plist for Languages?

IIRC @concertman told me iOS doesn't have a JSON parser in the standard library. In order to make Gherkin3 a zero dependency library on iOS we'd have to read in  the `gherkin-languages.json` in a natively supported format. Would it make sense to generate a `gherkin-languages.plist` file during the build that can be read more effortlessly `GHGherkinDialectProvider`? We could use a scripting language for that.

## Ignoring files

Can you add a `.gitignore` please?

LiohAu:
Thanks. I read that less 1 week was needed somewhere, I tried to respect the rule :P

Build Artefacts : I configured the build directory in Xcode and I thought that it was saved in the .xcodeproj file, but it appears that I was wrong. So i will update the Makefile with the CONFIGURATION_BUILD_DIR variable and it should works.

OS: The Gherkin code is working for iOS and OSX, there are 7 build targets configured in the .xcodeproj : 
1/ GherkiniOS => iOS library static which should produce the libGherkiniOS.a
2/ GherkinTestsiOS => Unit test library for libGherkiniOS.a, maybe we should remove this one, except if we want to check the unit tests works on iOS devices like they do on Macs. And without a connected iOS device, we will execute test through a simulator (so on the OSX cpu.. but with the iOS SDK) - Not really tested yet.
3/ AstGenerator = Build the AstGenerator app.
4/ TokensGenerator = Build the TokensGenerator app.
5/ GherkinLanguages = Create a .bundle file containing the gherkin-languages.json (I can skip this bundle and the target if we generate a const NSString * with the content of the file during the make process)
6/ GherkinOSX = OSX static library for OSX which should produce the libGherkinOSX.a
7/ GherkinTestsOSX = Unit test library for libGherkinOSX.a - Not really tested yet.

Currently the Makefile, only builds the "GherkinOSX", "AstGenerator" and "TokensGenerator" targets (and execution of the GherkinTestsOSX is commented) but you|we can easily add a new xcodebuild line for the iOS versions.

Plist: iOS / OSX did not have a JSON parser few years ago, but now there is the NSJSONSerialization class that I use in the AstGenerator for example.

Ignoring files: Yes, I should have done it since the beginning..

LiohAu:
I just fixed the build folder issue, and created the .gitignore file.

LiohAu:
All the acceptance tests seems to be ok :)

aslakhellesoy:
Looks great @LiohAu - will try to merge this in tomorrow!

PR #111, opened 2015-10-19
Markdown support (WIP_

aslakhellesoy:


stof:
should a markdown format for cucumber pickles really belong to the cucumber/gherkin repo ? Parsing pickles from markdown is not related to a gherkin parser (it is not the gherkin language anymore).
This makes me think that pickles themselves should probably belong to a separate package, to be reusable. And then, you would have the gherkin compiler depending on this package to convert the gherkin AST to pickles. And another package could depend on the pickles as well but build it from a markdown AST.

aslakhellesoy:
You're right @stof - both pickles and a markdown parser would ideally belong in a different project. However - many small repos causes a lot of overhead due to the multi-platform nature of this library, so at the moment it's more pragmatic to keep it all in the same repo.

stof:
Well, you could then keep a single big repo for the different things, but organize language folders to have different packages in them (and separate subtree splits). This would allow you to have the main repo to develop things, but still generate several packages for each language.

stof:
btw, you might be interested in the splitsh tool to manage the subtree split. Part of it has been open-sourced already at https://github.com/splitsh/lite, and you could ask for usage of the full version of the tool which is not yet ready for open-sourcing (see the end of the readme)

aslakhellesoy:
@stof - thanks a lot for the pointer to `splitish-lite` and the wonderful talks and articles I found via that.

I'm pretty confident a monorepo for the entire project makes sense - at least I'd like to try it. We'll see how this goes!

https://twitter.com/cucumberbdd/status/745379183421886464

PR #114, opened 2015-10-25
Escape only '|', 'n' and '\' in table cells.

### LINKED ISSUE #113: Allow regexp in table cells
Like `username:\s+xyz` - see comments in [this blog post](https://cucumber.io/blog/2015/09/11/cucumber-2.1)

Might be relevant to #40 and #71

brasmusson:
From what I see in Gherkin2 (explored through Cucumber-JVM), the behaviour of '\' in table cell is:
```
'\n' -> <newline>
'\|' -> '|'
'\\' -> '\'
\<other>' -> '\<other>'
```

This PR makes Gherkin3 behave the same, with the following addition:
`'\\n' -> '\n'` (the character '\' followed by the character 'n'). 
In Gherkin2 both `'\\n'` and `'\\\n'` becomes `'\<newline>'` which makes it impossible to get a string with the character '\' followed by the character 'n' from a table cell.

The Objective-C fix is untested, since I (so far) have no Objective-C compiler on my Linux box.

Fixes #113.


akostadinov:
just FYI, changing escape rules **again** will:
* break compatibility for users that already escape characters like `\\s` in regular expressions
* will prevent backward compatible escaping of characters other than `\n` and `\` and `|`

I would prefer to see cucumber keep `\` always as an escape and add in the future more characters to be escaped like `\t`, numeric ascii and perhaps random unicode code. Perhaps even forbid escaped characters that do not have special meaning to prevent users from using strings that may change behaviour at some point. e.g. if user puts in a table `\s` raise an error because `\s` is not s special escape and if in the future such escape is defined, then behaviour may change.

In any case taking a decision on that and sticking to it is most important as introducing syntax incompatibilities is highly disruptive.

brasmusson:
Is this PR changing the escape rules again, or is it fixing the bug that made the behaviour of `\` in table cells different from Gherkin2?

akostadinov:
@brasmusson , I suspect this is meant as a rhetoric question but I'll answer. From my point of view it is changing behaviour as there've been a good time now all backslashes `\` have been handled differently and now the *fixing* is an actual change in behaviour for existing users (whether or not original feature was meant like that or not).

brasmusson:
@akostadinov Yes, the question was mostly meant as a rhetoric question, but afterwards it occurred to me:
> break compatibility for users that already escape characters like `\\s` in regular expressions

No, also after this change `\\s` will work since `\\` is (still) changed to `\`, that means that after this change both `\\s` and `\s` will produce `\s`.

akostadinov:
@brasmusson, ha, you're right.
I would have preferred a couple of more escapes but as long as consensus goes with only 3, I can live with it.

Thanks.

PR #115, opened 2015-11-01
Do not change escaped docstring separators in descriptions.

brasmusson:
Now the unescaping of docstring separators is performed unconditionally when matching the token "other". Since the token "other" is used both for descriptions and docstrings, also in descriptions `\"\"\"` is replaced with `"""` (why anyone would write `\"\"\"` in a description is another question).

This PR changes the behaviour so that only in docstrings  `\"\"\"` is replaced with `"""`.

### DIFF
--- a/dotnet/Gherkin/TokenMatcher.cs
+++ b/dotnet/Gherkin/TokenMatcher.cs
@@ -226,7 +226,7 @@ namespace Gherkin
 
         private string UnescapeDocString(string text)
         {
-            return text.Replace("\\\"\\\"\\\"", "\"\"\"");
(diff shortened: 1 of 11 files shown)


aslakhellesoy:
LGTM! Feel free to merge @brasmusson

brasmusson:
@aslakhellesoy Now this PR is merged, but I could not `push-subtrees` to gherkin-objective-c, I got an `ERROR: Permission to cucumber/gherkin-objective-c.git denied`.

aslakhellesoy:
You should have karma now @brasmusson

brasmusson:
That's right, thanks.

PR #121, opened 2015-11-05
Prevent token_scanner.py from (recklessly) monkey-patching io.StringIO.

zbmott:
By monkey-patching `io.StringIO` to be `cStringIO.StringIO`, `token_scanner.py` was inadvertently breaking breaking any code that expected `io.StringIO` to be of type `type`, because `cStringIO.StringIO` has type `builtin_function_or_method`.

Practically speaking, this means that any code that attempted to interact with `io.StringIO` as one would interact with a class (e.g. subclassing, accessing a class attribute, etc) would fail with an error, for example:
* Subclassing: `TypeError: Error when calling the metaclass bases`
* Attribute access: `AttributeError: 'builtin_function_or_method' object has no attribute 'write'`

The [Celery](https://github.com/celery/celery/) project does exactly this in the `celery.five` [module](https://github.com/celery/celery/blob/v3.1.16/celery/five.py#L383-L387), which is how I encountered this error.

### DIFF
--- a/python/gherkin3/token_scanner.py
+++ b/python/gherkin3/token_scanner.py
@@ -2,11 +2,11 @@ import io
 import os
 from .token import Token
 from .gherkin_line import GherkinLine
+
 try:
     from cStringIO import StringIO
-    io.StringIO = StringIO
 except ImportError:
-    pass
+    StringIO = io.StringIO
 
 
 class TokenScanner(object):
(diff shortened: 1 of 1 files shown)


brasmusson:
Thanks. At the end of the day I reverted back to the state before the "reckless" monkey-patching of io.StringIO and used `from ... import ...` for both Python2 and Python3 (https://github.com/cucumber/gherkin3/commit/d6c368d9a499ec526f5ecab9bad15779f298a264) - except for the use of sys.version_info to detect the Python version (which has been used in other parts for the code).

PR #128, opened 2015-11-12
Fix #127: Ruby build error

### LINKED ISSUE #127: Fix ruby build
Looking at [this build](https://travis-ci.org/cucumber/gherkin-ruby/builds/90416555) it looks like something (simplecov?) depends on some junk called tins that only runs on ruby 2.0.

Let's just remove simplecov then.

Zearin:
Let’s see if a simple version cap in `.travis.yml` can fix #127.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,8 @@ matrix:
       rvm: jruby-1.7.22
       before_install:
         - gem install bundler -v 1.7.9
+        # tins >= 1.7 requires Ruby 2
+        - gem install tins -v 1.6 
 
     - env: TARGET_LANG=javascript
       language: javascript


Zearin:
w00t!  Ruby build is fixed!

brasmusson:
Thanks for your effort. I am not sure which is the best solution, but in the end I choose to fix the JRuby build by capping the `coveralls` version at 0.8.7 in the `gherkin3.gemspec` (https://github.com/cucumber/gherkin3/commit/a5fe2d77b57dc7da894860cbdbd2f1ac8a366186), then people using Ruby version >= 1.9.3 and < 2.0 are still able to do `bundle update` using the `gherkin3.gemspec`.

PR #134, opened 2015-11-18
Better French i18n

moreau-nicolas:
Hello!

As a native French speaker I was amazed to see that Gherkin supports writing features in my mother tongue. But I was a bit bothered by the fact that some expressions that I thought should be keywords weren't, or didn't conform to the French grammar.
This PR aims at fixing those grammar issues.

In the process of making those fixes, I spotted an issue with the python pickles compiler regarding UTF-8 encoding and fixed it.

I did run the full test suite with
```
make clean all
```
but couldn't run the objective-c suite because my Linux box doesn't have Xcode... (I don't own an Apple computer and don't have easy access to one).
All the other tests pass!


brasmusson:
Actually there were some more issues with the Python pickles compiler regarding UTF-8 encoding - related to Scenario Outlines in this case. Now they are also fixed https://github.com/cucumber/gherkin3/commit/46b79108cfabba11999c39d0a12ec5561de5e632.

aslakhellesoy:
Hi @moreau-nicolas,

An identical change was suggested a couple of years ago by @gregorybesson in cucumber/gherkin#223. I ended up rejecting it :crying_cat_face: 

Reading your arguments and thinking about it I'm happy with this change.

Merci!

gregorybesson:
Hello @aslakhellesoy !
It's never too late :) 
Congratz @moreau-nicolas : I think this is a great addition. Maybe I was rite too early :D 

I'll be happy to benefit from this addition.
Greg

moreau-nicolas:
@aslakhellesoy You're welcome, I'm happy to help enhance an already great tool. :smiley: 
@brasmusson, @gregorybesson Merci!

brasmusson:
Hmm, I was unaware of the history of related, rejected PR's in gherkin2, and honestly mostly focused in getting in the in the fixes in the Python compiler here - and the related ones I tracked down. They were actually merged into the master branch separately. Rather than leaving this PR half merged, I merged all of it. I'm glad everything ended up OK.

brasmusson:
I've come to realize that this change actually introduce kind of a backward compatibility problem. An existing feature with a step:
```
  Etant donné que j'aime les gâteaux
```
and a step definition with the regex:
```
  /^que j'aime les gâteaux$/
```
did work before but will not work after this change, since now the `que` will be considered as part of the keyword and needs to be removed from the step definition regex. This raises the question of the numbering of the next gherkin3 release (and the next Cucumber-Ruby release).

aslakhellesoy:
I've been a little worried about this too. It's true this would be a backwards incompatible change. It's weird to have gherkin3 version 4 though. I'd be happier to have gherkin version 4 if you see what I mean.

Maybe it's time to rename this repository (and the various packages) from gherkin3 to simply gherkin. And rename the old repo to gherkin2.

WDYT @brasmusson?

jbpros:
+1 to *gherkin* only. Having a version number in the name makes it confusing and hard to stick to semver. When following the semver rules, the major version number can climb quite quickly, often too quickly to have it included in the package name itself.

aslakhellesoy:
The problem is package names. I don't think the name gherkin is available on all platforms. Python I think.

brasmusson:
IIRC one of the reasons of naming the ruby-gem from this project "gherkin3" was to enable the use of both the gherkin(2) gem and the gherkin3 gem, the first spike in Cucumber-Ruby did use both, didn't it. We have passed that for the ruby version. It is a matter of timing, renaming the repo now when Cucumber-JVM still used gherkin2 might lead to that some issues that should be raised on that repo is raised here instead (they wont be fixed either way). Renaming the released artifacts would block use of the parser from here, and for instance the tag expression handling for command lines from gherkin2. But where would that be a problem? In Cucumber-JVM? In Cucumber-JS?

aslakhellesoy:
Yes, ruby is not a problem any longer. Java and JavaScript should be fine too.

The main problem is that gherkin is already a thing on pypi: https://pypi.python.org/pypi/gherkin

PR #135, opened 2015-11-19
Add gnéchlár for feature for irish

merrua:
Gnéchlár is a more correct word for a software feature in irish.

### DIFF
--- a/dotnet/Gherkin/gherkin-languages.json
+++ b/dotnet/Gherkin/gherkin-languages.json
@@ -1023,7 +1023,8 @@
       "Samplaí"
     ],
     "feature": [
-      "Gné"
+      "Gné",
+      "Ghnéchlár"
     ],
     "given": [
       "* ",


moreau-nicolas:
Hi @merrua!
I think you should commit the generated files as well (like I did in 42e733e).
And perhaps add some tests?
Thanks in advance!

aslakhellesoy:
@moreau-nicolas what sort of bugs do you think tests for a translation would catch?

moreau-nicolas:
I can think of a few things:
* encoding problems (I wouldn't have spotted the python problem I fixed in #134 if I hadn't added a non-ASCII test feature)
* duplication issues (tests helped me spot the problem that @brasmusson fixed in #138)

But those problems have been fixed and shouldn't happen again, so I agree tests aren't necessary for @merrua's proposal. Habbits die hard, sorry!

merrua:
Sorry I'm late noticing this. Is the conclusion you want tests or you don't want tests for this? I don't have the stack set up at the moment.

moreau-nicolas:
Hello! I no longer think that test are necessary for this patch.
I do think you should run `make update-gherkin-languages`, commit and update the pull-request, though.

Issue #153, opened 2016-02-04
Bring it closer to gherkin2 grammar


aslakhellesoy:
See details here: https://groups.google.com/forum/m/#!msg/cukes/SdEJpBG4twQ/EnIRAhp8EgAJ

mattwynne:
So @enkessler can you tell us some specific examples of what's missing? Your example in the google group seemed to refer to a `Scenario:` with no steps...?

aslakhellesoy:
Actually, individual tickets for each incompatibility would be ideal (with an example to reproduce). Then we can close this ticket.

enkessler:
I suppose that I could dig out the examples that I use in my own test suites and probably come up with a few more but, if you want to get back to the gherkin2 grammar, why not simply grab the grammar out of source control and reuse it? Is it not in handy form like the current grammar is and would have to be reverse engineered? I seem to recall state machines being involved.

brasmusson:
We do not want to get back to the gherkin2 grammar. One deliberate deviation which comes to mind (which is backward compatible BTW) is to allow comment anywhere in the feature file (which is why they in the new AST all are associated to the feature).

enkessler:
Oh, yeah. I noticed that. Now comments at the end of a file won't get lost for want of some object to attach them to like they used to. :)

So would it be fair to say that you are aiming not for something similar to the previous grammar but for something that is more flexible in general? After all, the gherkin gem can likely parse any reasonably sane feature file into something meaningful. And any upstream consumer can decide what is actionable or not (e.g. Cucumber can decide that a stepless test doesn't need to be run).

I can crank out an issue for each feature file that is potentially 'valid'. Some of them may require an update to the current grammar, some of them may already work and could just use a test (I only saw 6 tests while I was working on #155), and some of them may be thrown out entirely depending on where the line of 'usefulness' is drawn.

aslakhellesoy:
That would be very helpful. Keep in mind that most tests are cross platform and live under `/testdata`. They are run with `make`

enkessler:
I found them. I can't run them yet because, you know, Windows user and all that, but the tests look like straightforward input/output combinations. So, I get the gist of it.

Funny thing: several of the examples 'bad' input files are exactly the kind of use cases that I would be suggesting.

[...]

PR #160, opened 2016-02-14
Allow incomplete scenario outlines

brasmusson:
This PR change the Gherkin parser to not consider an incomplete scenario outline an error. This fixes one aspect of #153.

- [x] .NET
- [x] Go
- [x] Java
- [x] Javascript
- [x] Objective-C
- [x] Perl
- [X] Python
- [x] Ruby


enkessler:
Nice job tackling the thing that has the most use cases first!

The only other case that I can think of that I don't see in your commit is an outline with steps but not an example block.

    Scenario Outline: only steps
      * a step

aslakhellesoy:
This looks good to me - let's go ahead and implement it for all platforms!

aslakhellesoy:
Getting there!

brasmusson:
@aslakhellesoy I got `ERROR: Permission to cucumber/gherkin-perl.git denied` when pushing subtrees after having merged this PR. It seems like the access rights for the repository cucumber/gherkin-perl is not setup as expected.

aslakhellesoy:
Try again?

brasmusson:
Now it worked, thanks.

aslakhellesoy:
That is a phenomenal job @brasmusson. Thanks!!

SabotageAndi:
Any plan when a version with this PR is on NuGet.org?

aslakhellesoy:
@SabotageAndi we don't provide estimates, but you can look at the [CHANGELOG](https://github.com/cucumber/gherkin/blob/master/CHANGELOG.md) to get an idea of how frequently we have made releases in the past.

PR #162, opened 2016-02-23
enable empty examples table of a scenario

SabotageAndi:
This PR enables, that example tables could be empty.

In a tool for SpecFlow (SpecFlow+Excel, http://www.specflow.org/plus/excel/), you can get the examples from an external Excel file. You have to add a tag to the examples and the SpecFlow plugin generates the tests.
In this case the example tables are empty. The header row is for matching parameters and the columns in the Excel sheet.
As SpecFlow 2.0 switched to this Gherkin parser, we want to upgrade the tool to also use the new parser.

I talked with @gasparnagy about this behaviour and he thinks there was no particular reason why empty example tables should not be possible.

I hope the PR is complete and correct.
My changes where to the gherkin.berp and the incomplete_examples_table.feature test case.
If I missed something, I am happy to correct it.


brasmusson:
@SabotageAndi There is already a PR that allows empty example tables, and even less complete scenario outlines: #160. Right now I think we should build the behavior specified there (I did not implement more than one parser version until I got agreement on that behavior).

SabotageAndi:
@brasmusson Ok, I did not see the #160 PR. Let's continue there.
But why did you only implemented one parser? I thought they are all generated from the berp-file.
Did I missed something?

brasmusson:
@SabotageAndi Yes, all the parsers are generated from the berp-file. But the change in #160 does also require a change in other files like the AstBuilders (by introducing the Example_Table concept in the berp-file), so I made the complete change of one language - and did not even generate new parsers for the rest of the languages.

SabotageAndi:
@brasmusson Thanks for the explanation!

PR #163, opened 2016-02-26
Contextual errors (don't merge)

pjlsergeant:
Adds contextual errors eg:

```
-- UnexpectedEOF Error --

unexpected end of file, expected: #TableRow, #DocStringSeparator, #StepLine, #TagLine, #ExamplesLine, #Comment, #Empty
  at [../testdata/bad/unexpected_eof.feature] line 5, column 0

-- [../testdata/bad/unexpected_eof.feature] --

   1| Feature: Unexpected end of file
   2| 
   3|   Scenario Outline: minimalistic
   4|     Given the minimalism
   5| 
    +-^

----------------------------------------------
```

```
-- UnexpectedToken Error --

expected: #Language, #TagLine, #FeatureLine, #Comment, #Empty, got 'invalid line here'
  at [placeholder] line 2, column 1

-- [placeholder] --

   1| 
   2| invalid line here
    +-^
   3| 
   4| Feature: Single parser error
   5| 

-------------------
```

Are these what we want? Should they be the default? Thoughts please...

### DIFF
--- a/perl/bin/gherkin-generate-ast
+++ b/perl/bin/gherkin-generate-ast
@@ -17,6 +17,8 @@ sub run {
 
     my $parser = Gherkin::Parser->new();
 
+    local $ENV{'GHERKIN_VERBOSE_EXCEPTIONS'} = 1;
+
     for my $file (@file_list) {
         my $base;
         $base = $json->encode( $parser->parse($file) );
--- a/perl/lib/Gherkin/AstBuilder.pm
+++ b/perl/lib/Gherkin/AstBuilder.pm
@@ -68,15 +68,23 @@ sub get_location {
     use Carp qw/confess/;
     confess "What no token?" unless $token;
 
-    if ( !defined $column ) {
-        return $token->location;
-    } else {
-        return {
-            line    => $token->location->{'line'},
-            context => $token->location->{'context'},
-            column  => $column,
-        };
-    }
+    return {
+        line => $token->location->{'line'},
+        column => (
+            defined $column ? $column : $token->location->{'column'}),
+    };
+}
+
+sub get_error_context {
+    my ( $self, $token ) = @_;
+
+    use Carp qw/confess/;
+    confess "What no token?" unless $token;
+
+    return {
+        filename => $token->location->{'filename'},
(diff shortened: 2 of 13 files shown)


aslakhellesoy:
I might have misunderstood what you meant. I don't think we should pollute the AST with context nodes like this. Besides, for most parse errors we can't even produce an AST.

I thought the errors with context were just going to be printed to stderr alongside the existing errors.

pjlsergeant:
Leaving it in the AST was oversight on my part, now removed.

aslakhellesoy:
That's really nice!

Is this a "home made" error format or does it follow some existing convention for parse error reporting?

It looks a little fiddly to implement - would you be willing to contribute patches to any of the other 7 implementations? If so, which ones?

pjlsergeant:
Format is entirely home-made, and just trying to look similar to what I used for https://github.com/pjlsergeant/test-bdd-cucumber-perl

If there's some kind of standard we could use instead, that would probably be better. Perhaps an even better solution is to upgrade the exception classes (see: https://github.com/cucumber/gherkin/pull/164) to contain structured data, and then people can catch and format however they like. That might not be a bad first step to get done anyway.

The formatting code *is* currently fiddly, but I suspect that's more of a function of writing it in a taxi one morning, rather than the problem being inherently hard; I suspect a little bit of pre-thought on how it's done will give us something a lot simpler, that would be fairly easy to replicate in the other languages.

I have no problem at all contributing other implementations, but I am unusually time poor between two fulltime jobs, an unfinished MSc, a wife who does occasionally enjoy my company, and a travel schedule that means I will literally only see my flat for about 10 days in the next 90 :-D

If you like what you see, maybe we could start with me trying to put together a much cleaner version that would be far less fiddly for other implementors.

pjlsergeant:
Also, next on my todo list of Cucumber-related projects is a Perl module that provides a slow but featureful wrapper for the AST, so that people who actually want to use the parser have something a lot more userfriendly - basically implementing your class diagram as actual classes, with documentation. `Gherkin::Model` or `Gherkin::Simple` or something, which uses `Gherkin` as the underlying parser.

PR #174, opened 2016-03-13
Background as scenario definition

aslakhellesoy:
The main purpose of this change is to simplify client code:

* No need to check whether feature.background is null - just loop
* Handle Background the same loop as Scenario and ScenarioOutline

This change also simplifies gherkin itself - I was able to delete about 30 LOC total.

The main reason I want to change the AST API as proposed in this PR is that I'm working on a new HTML reporting library that could be integrated with any Gherkin 3 based Cucumber. Having it this way makes that reporting UI simpler and cleaner.

I haven't done Perl or Objective C yet - maybe @pjlsergeant, @Ahmed-Ali or @LiohAu can help with that? Or @brasmusson ?


aslakhellesoy:
The proposed change is backwards incompatible, so this would mean a major version bump to 4.0.0. The change is fairly small though.

Cucumber maintainers @tooky @mattwynne @jbpros @charlierudolph @Ahmed-Ali any thoughts/objections?

javascript/test/parser_test.js
```diff
@@ -47,13 +45,12 @@ describe('Parser', function () {
...
       name: 'hello again',
       description: undefined,
-      background: undefined,
       scenarioDefinitions: [],
       comments: []
     });
   });
 
-  it("parses feature after parse error", function () {
+  it.only("parses feature after parse error", function () {
```
charlierudolph: This `.only` needs to be removed.

javascript/test/parser_test.js
```diff
@@ -71,6 +68,7 @@ describe('Parser', function () {
 						    "      closed docstring\n" +
 						    "      \"\"\""),
 			   matcher);
+console.log(JSON.stringify(ast, null, 2))
```
charlierudolph: I would guess this should be removed to

charlierudolph:
After my comments are addressed, looks good from the JS point of view.

javascript/test/parser_test.js
aslakhellesoy: Thanks!

Ahmed-Ali: For the current Objectieve-c implementation; It is safe to execute methods
on nil (aka null) values by default; so it should be fine.
Also, the current implementation of Cucumberish executes backgrounds in the
same loop as the scenario and scenario outline already.

As a quick attempt, I have changed the parent of GHBackground to be
GHScenarioDefinition and seems to be fine; however, I need to do more test
for this.

So to conclude, it will not require too much work to achieve what is
mentioned in the main purpose.


Best Regards,
Ahmed Ali
Software Engineer specialized in Mobile Development
Phone: +31 062 0991 184

Email: eng.ahmed.ali.awad@gmail.com
LinkedIn: http://eg.linkedin.com/in/engahmedali

Ahmed-Ali:
For the current Objectieve-c implementation; It is safe to execute methods on nil (aka null) values by default; so it should be fine.
Also, the current implementation of Cucumberish executes backgrounds in the same loop as the scenario and scenario outline already.

As a quick attempt, I have changed the parent of GHBackground to be GHScenarioDefinition and seems to be fine; however, I need to do more test for this.

So to conclude, it will not require too much work to achieve what is mentioned in the main purpose.

mattwynne:
I think this change is a good idea, but I'm uncomfortable with the name `scenarioDefinitions` for this abstraction - it doesn't really seem to mean anything. Could we just call this collection `children` instead, since it contains the direct child nodes of a `Feature`?

aslakhellesoy:
Yeah I had the same thought - I think children is a better name. I can make that change.

PR #179, opened 2016-03-20
Add Rule synonym for Scenario Outline

aslakhellesoy:
Please make commits for more translations in languages you know! /cc @cucumber/core


enkessler:
I believe that pirates think of a 'rule' as a 'guideline'. ;)

PR #187, opened 2016-03-25
Gherkin parser and compiler in c.

brasmusson:
This is still work in progress.

The idea is that the Gherkin parser and compiler should have no dependencies, and with make and a c compiler it should be possible to build the library with the Gherkin parser and compiler, and to run the acceptance tests.

The things to do include (but are not necessarily limited to):
- [x] Add a Readme.md with basic instructions for the building and using of parser and compiler.
- [ ] wchar_t is used to store the unicode strings from the feature file. The size of wchar_t is platform dependent, so somewhere not all 31-bit unicode codepoints that can be stored in a utf-8 file will be possible to store in the wchar_t used on the platform (I guess that some platforms use 16 bits for wchar_t). This should the TokenScannar catch at runtime, and signal an error.
- [x] Move the public header files to the directory `include` (from the directory `src`)
- [ ] Add the prefix `GH_` to the public types and symbols? (to make it less likely with naming conflict)
- [ ] Add the file name prefix `gh_` to the source files?
- [x] Extract some utility functions (for instance for copying string the the constructors), to reduce duplication.
- [ ] Build also a shared library (for linking at runtime)
- [x] Also use clang on Travis.
- [ ] Cover important/complex parts of the code with (gtest) unit tests. For instance the replacement of placeholders in outline steps.
- [x] Handle all types of linebreaks (`\n`, `\r\n`, `\r`) properly in the TokenScanner.
- [x] Implement a TokenScanner to read from (unicode literal) strings as the parsers in other languages can.
- [ ] ~Support the `perf` performance benchmarking script.~


jlangr:
This link: https://groups.google.com/forum/#!msg/cukes/g7wEFEGldLI/xHe6pmDMqhAJ from 4 years ago suggests that this larger PR contains a fix for the problem we are seeing (tables are not shown in the console output).

Any chance you can post just that fix?

Thanks!

brasmusson:
@jlangr It has been so long that the link is obsolete (the Gherkin2 repository has been moved). Short answer, No, there will not be a new release of Gherkin2 (current version, Gherkin v2.12.2) with that problem fixed (as this repo is a complete rewrite of the Gherkin library - the current release is Gherkin v4.0.0 - which is not used by Cucumber-JVM yet). For a work around, see my (new) answer on the thread: https://groups.google.com/d/msg/cukes/g7wEFEGldLI/C3pbVl-FBQAJ

aslakhellesoy:
Do you think it would be hard to implement the cli interface that just landed on master? That could be used to build a portable C `gherkin` executable, which could be compiled on most platforms. That would be awesome!

brasmusson:
No it is not very hard to implement the cli interface, I tried and have done much of it already.

brasmusson:
Now does the c-version implements the events and the cli interface.

aslakhellesoy:
Exciting!!!

I'm getting a build error though:

```
gcc -c -Wall -Werror -g -MMD -MP -MF ../objs/ast_builder.d -I ../include -I . ast_builder.c -o ../objs/ast_builder.o
ast_builder.c:172:45: error: implicit conversion from enumeration type 'enum RuleType' to different enumeration type 'TokenType' (aka 'enum TokenType')
      [-Werror,-Wenum-conversion]
        token = AstNode_get_token(ast_node, Rule_BackgroundLine);
                ~~~~~~~~~~~~~~~~~           ^~~~~~~~~~~~~~~~~~~
1 error generated.
```

```
gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin15.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

I can get it to compile by removing `-Werror` in `src/Makefile`

However, it doesn't seem to parse correctly:

```
bin/gherkin_generate_tokens ../testdata/good/background.feature
(2:1)Empty://
(3:3)FeatureLine:Background/a simple background/
(4:1)Other:/    Given the minimalism inside a background/
(5:1)Other://
(6:1)Other://
(7:3)BackgroundLine:Scenario/minimalistic/
(8:5)StepLine:Given /the minimalism/
EOF
```

The expected output is:

```
(1:1)FeatureLine:Feature/Background/
(2:1)Empty://
(3:3)BackgroundLine:Background/a simple background/
(4:5)StepLine:Given /the minimalism inside a background/
(5:1)Empty://
(6:1)Empty://
(7:3)ScenarioLine:Scenario/minimalistic/
(8:5)StepLine:Given /the minimalism/
EOF
```

Quite different.

brasmusson:
Interesting. It does work on
```
$ gcc --version
gcc (Ubuntu/Linaro 4.6.3-1ubuntu5) 4.6.3
Copyright (C) 2011 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
([Travis](https://travis-ci.org/cucumber/gherkin/jobs/194020767)), it does work on
```
$ gcc --version
gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4
Copyright (C) 2013 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
(my machine), it does not work on
```
gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin15.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```
(your machine).

aslakhellesoy:
I think I've found out how to reproduce the build error on my local machine.

Run `make clean && make`. This fails for me (as described above). The build has changed `src/dialect.c` (see below). I can get the build to run with `git checkout src/dialect.c && make`.

There seems to be something wrong with the make dependency rules for `src/dialect.c` since it doesn't appear to be regenerated in CI.

Any idea why regenerating the file causes the scanner to misbehave?

[dialect.diff.txt](https://github.com/cucumber/gherkin/files/726067/dialect.diff.txt)

brasmusson:
There is of course the general weakness of (GNU) make - it uses time stamps to determine whether a rule shall be applied. That does not always work well with version control. Changing the branch could mean that the target file of a make rule is out of sync with the dependency files, but the time stamp on the target file is newer then all dependency files. Then the make rule is not applied, even though applying the rule would change the target file.

I did submit #210 to make the language update more robust in such situations (`make update-gherkin-languages`). The approach is (when the language file is not just copied - for instance for Go) to regenerate the file under a new name, and if the new file is different, then overwrite the old file, otherwise do not touch the old file (to avoid that make rules with it as a dependency are applied when the file content has not changed).

When I added the language to the pickles (#251) `make clean` followed by `make` reviled (again) that some files were out of sync, thus the first commit on that PR.

This PR has the more robust implementation for `make update-gherkin-languages`, but with just `make`, `src/dialect.c` is only regenerated if its time stamp is older than `../gherkin-languages.json` or `dialect.c.jq` (and newer time stamp does not guarantee it is not out of sync).

aslakhellesoy:
I know there is room for improvement in the targets that update the dialects. I have an idea about how to make it more robust that is a bit different from what you're suggesting, but let's ignore that for now. 

A safe way to regenerate the dialects is as follows (from the `c` folder):

```
git clean -dfx
make clean
make .built
git status
```

Do you see the same as me - `src/dialect.c` has been modified?

And do you get the following (incorrect) output for `bin/gherkin_generate_tokens ../testdata/good/background.feature`?

```
(2:1)Empty://
(3:3)FeatureLine:Background/a simple background/
(4:1)Other:/    Given the minimalism inside a background/
(5:1)Other://
(6:1)Other://
(7:3)BackgroundLine:Scenario/minimalistic/
(8:5)StepLine:Given /the minimalism/
EOF
```

Any idea what's causing this?

brasmusson:
> Do you see the same as me - src/dialect.c has been modified?

No. On Linux (with jq v1.4) the generation if `src/dialect.c` works as intended. It is only on Mac (with jq v.1.5) that the generation of `src/dialect.c` does not work properly, see [Travis](https://travis-ci.org/cucumber/gherkin/builds/195895970).

brasmusson:
It turns out that jq v1.4 and jq v1.5 behaves slightly different with respect to "iteration over lists", now the [jq filter](https://github.com/cucumber/gherkin/pull/187/commits/eb0575e088e9dd8bdcd3ee4e16689a8959a302ea) has been updated to produce the same result for both versions, so the [build passes](https://travis-ci.org/cucumber/gherkin/builds/195960881) for all platforms and jq versions.

aslakhellesoy:
Is this ready to be merged now @brasmusson?

brasmusson:
I think this is ready to be merged with the intended use case of parsing feature files and producing Gherkin events. To do that users only needs to:
* build the gherkin binary
* execute the binary and pipe the events to something that can consume them.

This is what the minimal README.md talks about. 

To integrate the gherkin library into something larger, a full C/C++ Cucumber, I could imagine requests for some things that are not available out of the box, such as:
* the ability to built a shared library (only a static library is build by the current makefiles),
* being able to integrate the building of the gherkin library with the build system of the larger thing (cmake or whatever)
* better documentation of the used data structures for the AST and Pickles,
* C++-wrappers to the AST and Pickles for simplified manipulation and memory management.

aslakhellesoy:
Thank you again @brasmusson for this monumental piece of work. Maybe one day this is the only parser we need?

PR #189, opened 2016-03-29
Allow empty Feature files

aslakhellesoy:
This change would allow empty feature files, or files with only comments. It changes the top level node of the AST to be a `FeatureFile`, and the `comments` array will move from `Feature` to `FeatureFile`.

See the original discussion in https://github.com/cucumber/gherkin/issues/153#issuecomment-201936912

What do you think @cucumber/gherkin - is this worthwhile pursuing? (Currently I've only modified the ruby AstBuilder to get an idea of how much work this is. Seems easy enough to do).


Zearin:
Makes sense to me.  Cucumber can worry about what to do with empty (or comment-only) feature files, but IMO Gherkin should just accept `*.feature` files as input and produce the parsed output.

brasmusson:
I'm not opposing the change itself, but isn't this something the triggers a major version change? Having the parser not return a Feature object, but an object that may contain a Feature object.

enkessler:
It's not backwards compatible so, yeah, it'd be a 'major' release.

aslakhellesoy:
@brasmusson the next release will be major anyway (4.0.0) because of #178 and d0e75802bdf40a7e515a5749c6632a524d559829 (which has a poor commit message, I know)

enkessler:
In that case, does it behoove us to buckle down and finish #153 in order to avoid yet another significant release after this one?

enkessler:
Did you have to do any of those parser changes by hand or is that purely the result of regenerating the parser with the new grammar file? I ask because I was going to get some of the other languages passing again but every regenerated parser seems to not produce an AST at all anymore.

aslakhellesoy:
@enkessler all of the parser files in gherkin are generated by berp. This is explained in more detail in the README.

enkessler:
@aslakhellesoy Yeah, I got that. But none of the other languages seem to work with their newly generated parsers and the manner in which they fail to work looks like it has to do with the parser itself. It would be odd if Ruby's new parser was the only one that didn't generate incorrectly, thus I was curious if you had to tweak it after the fact.

I see that you also changed the AST builder for Ruby but playing with the builder for the other languages didn't seem to solve the issue, so I wanted to double check that the parser itself wasn't the problem.

brasmusson:
@enkessler For this particular grammar change some of the "parser generation control files" (.razor) needs to be updated, since they (unnecessarily) have hard-coded assumptions about the top most grammar rule (see my change in the file python/gherkin-python.razor in 2ff68ee). As far as I see it is needed also for Javascript and Go. Otherwise it should just be to update the AST builder, compiler and maybe the unit tests - in addition to the regeneration of the parser from the grammar.

enkessler:
@brasmusson Thanks! I'll take another crack at the other languages.

Also, #194 has been waiting for few days now and it looks like you redid most of it yourself. Hooray duplicated effort!

enkessler:
Got JavaScript passing again in #194.

@cucumber/gherkin Shall we go ahead and have a 'type' of 'FeatureFile' at the top level of the AST in the same manner as the other objects have?

aslakhellesoy:
Yes I think that mkes sense. Since the source might come from a different source than a file I think `GherkinDocument` is probably a better name for the top node.

brasmusson:
If the `type` of the top level AST node is `GherkinDocument`, should also the grammar rule change to `GherkinDocument! := Feature?` ?

aslakhellesoy:
I've already done that in the grammar in this PR. Apart from the rename.

brasmusson:
My questing was about the rename (`FeatureFile` -> `GherkinDocument` in `gherkin.berp`), and I see that the answer is Yes.

brasmusson:
I have changed the top rule name, added the type attribute to the `GherkinDocument` AST node, and cherry-picked in the Javascript update from #194. While at it I also rebased to the tip of the master and squashed some commit to simplify the list of commit. I hope I did not cause anyone to much trouble by force pushing to the PR branch.

enkessler:
@brasmusson I decided to simply merge the upstream rather than rebase. I knew how to do it and it was a fun little code review. #194 is once again in sync with this PR, although there isn't much left in it. I'm giving up on Go and Java for the moment and trying to do Perl. I will go ahead and commit the razor file fix for Go, however.

enkessler:
@cucumber/gherkin Perl is now updated in #194.

Side note: Perl had some of its parser logic hard coded inside of [a separate class](https://github.com/cucumber/gherkin/blob/master/perl/lib/Gherkin/ParserBase.pm) instead of being derived from the grammar and its razor file. Can the razor file be adjusted to also generate that class? Should that class be tossed? Do we simply put up with potentially having to tweak it by hand every time the parser is regenerated?

brasmusson:
With another razor file and another make rule it is possible to generate another file besides the parser implementation, it is already done for [Objective-C](https://github.com/cucumber/gherkin/blob/master/objective-c/Makefile#L47-L51).

aslakhellesoy:
We shouldn't have hand-coded changes to generated parser code.

Further, `make clean` should remove all generated code.

aslakhellesoy:
Wouldn't it be sufficient to move the `parse` method from `ParserBase.pm` to `gherkin-perl.razor` and replace `Feature` with `RuleType.@Model.RuleSet.StartRule.Name`? (I guess this should be done in all implementations).

brasmusson:
The Java update is now fixed in this PR.

enkessler:
That just leaves Go, .NET, and Objective-C, I believe.

brasmusson:
The .NET update is now fixed on this PR, so now only Go and Objective-C remains.

enkessler:
@brasmusson Well, I don't have a machine that can do Objective-C and Go has been fighting me for the last day to no avail, so it's all you at this point. Just a reminder, I have already fixed the razor file for Go in my #194.

brasmusson:
Now only Objective-C remains. Also the Perl and parser/razor file changes for Go have been cherry-pick to this PR from #194.

I also move the Perl `parse` function from `ParserBase.pm` and made it generated from the razor file by berp.

Ahmed-Ali:
I will take care for the Objective-C implementation


Best Regards,
Ahmed Ali
Software Engineer specialized in Mobile Development
Phone: +31 062 0991 184

Email: eng.ahmed.ali.awad@gmail.com
LinkedIn: http://eg.linkedin.com/in/engahmedali

muhqu:
I don't get the benefit of putting the _Feature_ Ast Node inside top level _GherkinDocument_. Especially as there is only one single Feature allowed per document.

Of course I also want to allow empty feature files, but why not just return an empty feature object?
``` json
{
  "comments": [],
  "children": [],
  "keyword": "",
  "language": "en",
  "location": {
    "column": 1,
    "line": 1
  },
  "name": "",
  "tags": [],
  "type": "Feature"
}
```

I think it feels natural when you parse a `*.feature` file to get back a _Feature_ object… why do you want to change that?

Ahmed-Ali:
The Objective-C implementation is adopted now. However, will try to check the needed changes to the razor file later.

java/src/main/java/gherkin/ast/GherkinDocument.java
```diff
@@ -0,0 +1,25 @@
+package gherkin.ast;
+
+import java.util.Collections;
+import java.util.List;
+
+public class GherkinDocument extends Node {
```
brasmusson: Extending Node is a little bit of an anomaly since `GherkinDocument` does not have a location as all other `Node`:s (thus the `super(null)` to set the location to null). I did not experiment with not having `GherkinDocument` extending `Node` (and explicitly declare a `type` field in `GherkinDocument`), or if that would not work, move the `Location` field from `Node` to an intermediate class between `Node` and all its sub classes that have a location.

aslakhellesoy:
Great teamwork here folks! Is this mostly ready now? I'll take a stab at merging today and release 4.0.0

brasmusson:
I would say it is mostly ready now. I think that the two things to look at is the Objective-C implementation (still failing on Travis and there is a [comment](https://github.com/cucumber/gherkin/pull/189#issuecomment-207133642) about remaining work) and [my comment on the Java implementation](https://github.com/cucumber/gherkin/pull/189#discussion_r58982639).

aslakhellesoy:
@muhqu what you suggest could have worked, but I think this solution is a bit cleaner. 

Gherkin parses Gherkin documents, which often happen to be stored in .feature files, but could come from anywhere.

It's a design I borrowed from Java's .class files, which use a similar design - they have a CompilationUnit as the top level AST node, then one or more Class nodes underneath.

muhqu:
@aslakhellesoy 
> It's a design I borrowed from Java's .class files, which use a similar design - they have a CompilationUnit as the top level AST node, then one or more Class nodes underneath.

That makes sense as you can have multiple classes per file. But that should lead to the conclusion that a GherkinDocument may have zero or more Features, even though the common use-case is to have a single feature per file. Or am I missing something?

aslakhellesoy:
> But that should lead to the conclusion that a GherkinDocument may have zero or more Features, even though the common use-case is to have a single feature per file.

That's right - a `GherkinDocument` without a `Feature` keyword will have 0 features in it. I think that's better represented as a `feature` property that is null rather than a top-level `Feature` node with null values for keyword, name and description.

muhqu:
@aslakhellesoy but why not just go a step further and allow a GherkinDocument to have 0-N Features, at least from the AST point of view?

Don't get me wrong, I also like this effort here and also understand you want to ship a 4.0… but as it's already a breaking change you might want to think more than twice about how the AST should be structured...

aslakhellesoy:
@muhqu Oh I see your point about 0..N. That would be a good idea I think. But that can go into a later major release. Doing the rounds of 8-9 implementations takes a lot of time, and it's time to get a release out.

aslakhellesoy:
@Ahmed-Ali thanks for your change. Unfortunately, it doesn't seem to work.

What changed in all the other implementations is that the AST node returned by the parser is now a `GherkinDocument` instead of a `Feature`. (A `GherkinDocument` has a `feature` property that may or may not be null).

What seems to be missing from the Objective-C implementation is a `GHGherkinDocument` AST node class/interface. Also, the `GHParser::parse` method needs to return a `GHGherkinDocument`.

Let me know if you have any questions, or if you need help running the tests (they currently fail for objective-c).

aslakhellesoy:
See a7ecda01200f1248af99ea04bf4c4b45db213fc5 for an example of what changes were made for Java. /cc @Ahmed-Ali

Ahmed-Ali:
I will check it today and update the Objective-C implementation to match
these changes.


Best Regards,
Ahmed Ali
Software Engineer specialized in Mobile Development
Phone: +31 062 0991 184

Email: eng.ahmed.ali.awad@gmail.com
LinkedIn: http://eg.linkedin.com/in/engahmedali

objective-c/Gherkin/GHHasTagsProtocol.h
```diff
@@ -1,4 +1,5 @@
-﻿@class GHTag;
+﻿#import <Foundation/Foundation.h>
```
aslakhellesoy: Are all of these `#import <Foundation/Foundation.h>` changes needed? Would it be possible to make a change with more minimal modifications - just the minimum necessary?

Ahmed-Ali:
I can remove them; but the alternative to having them in place is to do this import in a precompiled header file. The only issue with that, it requires changes in the test target setup when you want to use Gherkin. And this can be tricky and makes it harder to get started with Gherkin

enkessler:
Please do not make it harder to get started with Gherkin. Learning to dual boot my machine and install all of the different languages and tools needed in order to (still not completely) build the project is exciting enough already.

aslakhellesoy:
@enkessler I'm not sure exactly what you're referring to or what you are suggesting. Can you be more specific?

aslakhellesoy:
@Ahmed-Ali my point is - if you need to make those changes, please do that on a different branch/PR. It's unrelated to this issue (introducing a new AST node)

aslakhellesoy:
@enkessler never mind - I understand your comment now. Note to self: read slower!

enkessler:
@aslakhellesoy Yeah, it took several weeks of trial and error in my spare time before I could finally build this project. It is not quite at the '_git clone_ and then _make_' level of usability that would be really helpful to random helpers like myself.

maxmeyer:
@aslakhellesoy Sidenote: We had a similar discussion @ cucumber/aruba with @e2. We decided to include a Docker file - details are pending - to make it easier for others to get started with aruba. This one will include the minimum required dependencies. Maybe this can help here as well.

aslakhellesoy:
@Ahmed-Ali I fixed it in 92e65690c19924c6e51bbaf8128c815540c9aa49 so no need to do anything :-)

Ahmed-Ali:
@aslakhellesoy totally agree, I should have included these other changes in a different branch.
And thanks for fixing it!

PR #194, opened 2016-03-30
Tests and implementations for incomplete features

enkessler:
This is intended as a supplement to #189.

I have updated all of the shared test data and got the Ruby, JavaScript, and Perl implementations passing.


enkessler:
Everything needed has been cherry picked out of here.

PR #199, opened 2016-04-04
AppVeyor Build for dotnet to get pre-release NuGet feed

SabotageAndi:
As with @aslakhellesoy shortly discussed on gitter, this is an AppVeyor yml that enables pre-release NuGets via an AppVeyor build.
AppVeyor is hosting the feed, so no need of NuGet.org account credentials.

You have to add the project in AppVeyor as described here: https://www.appveyor.com/docs
As the yml- file already exists, it is only the click on the add button.

In the Settings\NuGet section you will find the feed Url. Please update the one in dotnet/README.md to your value.
Also update the urls of the badge.

I assumed that the next version will be 3.3.0. If this is not correct, please update the GherkinVersion in environment.


appveyor.yml
```diff
@@ -0,0 +1,31 @@
+version: build{build}
+
+
+environment:
+  GherkinVersion: "3.3.0"
```
aslakhellesoy: The next version is going to be 4.0.0, but we didn't know that until recently - we thought it was going to be 3.2.1 until we introduced a backwards-incompatible change.

We won't always know what the next version will be, but I guess it's best to assume it will be the last release +1 patch - then update minor or major if we know it will change.

Anyway - please change this to 4.0.0

appveyor.yml
```diff
@@ -0,0 +1,31 @@
...
+  project_feed: true
+  disable_publish_on_pr: true
+
+configuration: Release
+
+
+before_build:
+  - appveyor DownloadFile https://dist.nuget.org/win-x86-commandline/latest/nuget.exe
+  - appveyor DownloadFile https://raw.githubusercontent.com/appveyor/ci/master/scripts/nuget-restore.cmd
+  - nuget-restore dotnet\Gherkin.CSharp.sln
```
aslakhellesoy: I think it would make more sense to move this entire file to the `dotnet` directory, and have Appveyor build the [gherkin-dotnet](https://github.com/cucumber/gherkin-dotnet) project instead. (We regularly `git subtree push` individual modules to independent repositories).

WDYT?

dotnet/README.md
```diff
@@ -1,3 +1,6 @@
 [![Build Status](https://secure.travis-ci.org/cucumber/gherkin-dotnet.svg)](http://travis-ci.org/cucumber/gherkin-dotnet)
+[![Build status](https://ci.appveyor.com/api/projects/status/ka74of1ptpdfr6uv?svg=true)](https://ci.appveyor.com/project/SabotageAndi/gherkin)
```
aslakhellesoy: I'd prefer to make the AppVeyor build owned by myself. Is that OK with you?

brasmusson: There is a [cucumberbdd](https://github.com/cucumber/aruba/issues/287#issuecomment-121805433) AppVeyor account created for and used by aruba (unless you didn't know).

SabotageAndi: I never wanted that the stuff run in my AppVeyor account if this change is added. 
As noted in the PR- text, this Urls have to be changed to the real ones.

So when there is already an AppVeyor account, you only have to add the project.

appveyor.yml
SabotageAndi: The appveyor.yml file has to be in the root folder. So moving into the dotnet directory, you have to create the build for the gherkin-dotnet project.
It really depends how often you are pushing the changes to the other repos. If you are doing it only short before a release, it is then not a great help to update SpecFlow always to the latest version.

SabotageAndi:
Version is updated to 4.0.0. Or should it be another one as the release already happened?

aslakhellesoy:
The thing is - I don't know what the next release number will be. We have the same problem with java/pom.xml I suppose. Let's assume 4.0.1 for now and we'll bump it if we know we'll bump the minor or major number.

SabotageAndi:
@aslakhellesoy It is updated to 4.0.1.

PR #223, opened 2016-08-04
Add JIRA related terms for `Feature` and 'Scenario`

bazzel:
Our customer uses [JIRA](https://www.atlassian.com/software/jira). This tool uses different wording for `Feature` and `Scenario`. To avoid confusion when he reads and writes .FEATURE-files, JIRA terms are added to `gherkin-languages.json`.

### DIFF
--- a/gherkin-languages.json
+++ b/gherkin-languages.json
@@ -661,7 +661,9 @@
     ],
     "scenarioOutline": [
       "Scenario Outline",
-      "Scenario Template"
+      "Scenario Template",
+      "Story Outline",
+      "User Story Outline"
     ],
     "then": [
       "* ",


enkessler:
I kind of want to say that it would be more appropriate to make a new language for Jira (e.g. 'en-jira') rather than changing the baseline English version (i.e. 'en')

aslakhellesoy:
Sorry, won't accept this. Here is why:

* http://blog.mattwynne.net/2010/10/22/features-user-stories/comment-page-1/
* https://skillsmatter.com/skillscasts/5101-let-me-tell-you-a-story

bazzel:
I accept, but do you think @enkessler's suggestion makes a good candidate?

aslakhellesoy:
No I don't. People might be inclined to think that a scenario is the same as a story, or a feature is the same as an epic. They are very different things, not just by name, but also by meaning.

BTW, you're a fast reader!

enkessler:
@bazzel, @aslakhellesoy is right in his assessment of the road that this kind of thing will lead people down and, due to that, I can't fault him for not sanctioning this kind of addition the tool. However, I'm the sort of person that, should people really want to jump into holes, will at least advise on the best way to land so that they don't end up breaking their legs.

But please try to not make a habit of jumping into holes.

bazzel:
I appreciate both your answers, which make me realize we have to re-consider our usage of the terms which, in the end, lead to an improved codebase and therefore product.

PR #229, opened 2016-09-03
Pickle filters

aslakhellesoy:
I'm not 100% sure whether pickle filtering ought to live in the pickle library itself, or in Cucumber. I'm inclined to put it into pickles (which I'd like to extract to an independent [standard library](http://docs.cucumber.io/docs/standard-library.html)), because that would allow other libraries like Gherkin Lint to use it too. But maybe it's too early to say whether this is the [right abstraction](http://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction).

I'm proposing adding it here for now, for lack of a better place to put it. I only wrote it to get a feel for how the new [Tag Expressions](http://docs.cucumber.io/tag-expressions/) library feels to use, and it felt really simple! Ideally I'd put it in the Cucumber-JVM 2.x codebase, but that doesn't exist yet (except for a somewhat experimental design on my local machine).

Can we build other filters on this abstraction such as line filter and pattern filter? What about filtering scenarios based on other criteria such as error message? That would require a mechanism to attach metadata to pickles, perhaps during a Cucumber run. Would this be an acceptable design if we go for immutable data structures? Attaching metadata to a pickle would result in a new pickle being created, keeping the original unchanged.

Could we somehow align this metadata API with the [Cucumber Event Protocol](http://docs.cucumber.io/docs/architecture/event-protocol.html)?

Let me know what you think @cucumber/developers

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -45,6 +45,11 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>io.cucumber</groupId>
+            <artifactId>tag-expressions</artifactId>
+            <version>1.0.0</version>
+        </dependency>
+        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>4.12</version>
--- a/java/src/main/java/gherkin/pickles/PicklePredicate.java
+++ b/java/src/main/java/gherkin/pickles/PicklePredicate.java
@@ -1,0 +1,5 @@
+package gherkin.pickles;
+
+public interface PicklePredicate {
+    boolean test(Pickle pickle);
+}
--- a/java/src/main/java/gherkin/pickles/TagExpressionPredicate.java
+++ b/java/src/main/java/gherkin/pickles/TagExpressionPredicate.java
@@ -1,0 +1,24 @@
+package gherkin.pickles;
+
+import io.cucumber.tagexpressions.Expression;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class TagExpressionPredicate implements PicklePredicate {
+    private final Expression tagExpression;
+
+    public TagExpressionPredicate(Expression tagExpression) {
+        this.tagExpression = tagExpression;
+    }
+
+    @Override
+    public boolean test(Pickle pickle) {
+        List<String> tagNames = new ArrayList<>();
+        List<PickleTag> tags = pickle.getTags();
+        for (PickleTag tag : tags) {
+            tagNames.add(tag.getName());
+        }
+        return tagExpression.evaluate(tagNames);
+    }
+}
--- a/java/src/test/java/gherkin/pickles/TagExpressionPredicateTest.java
+++ b/java/src/test/java/gherkin/pickles/TagExpressionPredicateTest.java
@@ -1,0 +1,48 @@
+package gherkin.pickles;
+
+import gherkin.AstBuilder;
+import gherkin.Parser;
+import gherkin.ast.GherkinDocument;
+import gherkin.deps.com.google.gson.Gson;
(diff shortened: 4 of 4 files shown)


enkessler:
My gut says to leave filtering to the consumer. It might be able to back that up with actual arguments at a later time.

Perhaps because filtering the pickles after they have been produced seems like something extra tacked on that the consumer could easily do if that how they wanted to play with pickles and being fancy and allowing for pickle filtering during the process of creating them seems a bit shoe-horned in?

@aslakhellesoy, that's a nice article. Personally, at step number 6, I tend to immediately break abstraction A into abstraction A-1 and A-2 and replace all instances of A with whichever of the new ones is more appropriate. So I'll end up in the same place that the article suggests but without all the extra code and eventual rewinding. ;)

brasmusson:
As part of the work with upgrading Cucumber-JVM to use Gherkin v.4 (https://github.com/cucumber/cucumber-jvm/pull/1035), I had to implement pickle filtering based on [tags](https://github.com/cucumber/cucumber-jvm/blob/use-gherkin4/core/src/main/java/cucumber/runtime/TagPredicate.java), [line/location](https://github.com/cucumber/cucumber-jvm/blob/use-gherkin4/core/src/main/java/cucumber/runtime/LinePredicate.java) and [name](https://github.com/cucumber/cucumber-jvm/blob/use-gherkin4/core/src/main/java/cucumber/runtime/NamePredicate.java) - to replace the functionality of the `FilterFormatter` in Gherkin v2. It does not (of course) use the new tag expressions, but the [old tag expressions](https://github.com/cucumber/cucumber-jvm/blob/use-gherkin4/core/src/main/java/cucumber/runtime/TagExpression.java) from Gherkin v2.

mattwynne:
i think filters are about my favourite thing about the test case / pickle abstraction, and we've found [all kinds of pieces have fallen into place](https://github.com/cucumber/cucumber-ruby/tree/master/lib/cucumber/filters) when we started to use them in Cucumber Ruby.

I think it's fine putting them in here for now, but I agree that they ultimately belong in a different library. We've got some of our filters in cucumber-ruby and some in cucumber-ruby-core (which is our equivalent of pickles), so I'm not quite sure where's the best place either yet. I think the concept and abstraction / interface belongs in the pickles/core library, but maybe some of the implementations are for a client (e.g. we use them to add before hooks to test cases - pickles doesn't need to know or care about such things).

It's worth noting that we use a different API for filters in Ruby, though the general idea is the same. Ours allows filters to emit [modified copies of test cases](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/filters/apply_after_step_hooks.rb), and [change the order](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/filters/randomizer.rb), as well as [excluding some of them](https://github.com/cucumber/cucumber-ruby-core/blob/master/lib/cucumber/core/test/filters/locations_filter.rb).

mattwynne:
Oh and +1 for making pickles a standard library.

brasmusson:
>  Ideally I'd put it in the Cucumber-JVM 2.x codebase

The minimal effort way to get it into Cucumber-JVM 2.0 is IMHO neither to put it here nor in a standard library, but to implement in on top of cucumber/cucumber-jvm#1035, only a forth `PicklePredicate` class and some code to to choose between the new TagExpressions and the old style TagExpressions.

brasmusson:
Now the new tag expression syntax and the tag-expressions library part of cucumber/cucumber-jvm#1035, see https://github.com/cucumber/cucumber-jvm/pull/1035/commits/e14c9809174446ca31b91a5299acd0f8c90cbd5c.

PR #240, opened 2016-12-02
WIP: Cucumber Event Protocol / Stream API / Command-line interface

aslakhellesoy:
This PR adds a new streaming API for Gherkin, which outputs `source`, `gherkin-document` (AST) and `pickle` events according to the [Cucumber Event Protocol](https://docs.cucumber.io/event-protocol/).

It also has adds a CLI that makes it possible to use Gherkin as a standalone program. This is good news for Cucumber implementors, as it obviates the need to write a native Gherkin parser for a Cucumber implementation in a new language.

The following implementations have a complete streaming API and CLI:
* Java
* JavaScript
* Ruby
* Python

The following implementations are incomplete, and I need some help with these as I'm not fluent enough in these languages:

* [ ] Perl - @pjlsergeant ?
  * [ ] Print errors as ndjson
  * [ ] Replace `bin/gherkin-generate-ast` and `bin/gherkin-generate-pickles` with a new `bin/gherkin`
  * [ ] Implement stream API
* [ ] Objective-C - @LiohAu / @joshski / @Ahmed-Ali ? 
  * [ ] Print errors as ndjson
  * [ ] Replace `bin/gherkin-generate-ast` with a new `bin/gherkin`
  * [ ] Implement stream API
  * [ ] Implement a pickles compiler and use it in stream API and `bin/gherkin`
* [ ] Go - @muhqu / @koterpillar / @l3pp4rd ?
  * [ ] Print errors as ndjson
  * [ ] Replace `bin/gherkin-generate-ast` with a new `bin/gherkin`
  * [ ] Implement stream API
  * [ ] Implement a pickles compiler and use it in stream API and `bin/gherkin`
* [ ] .NET - @gasparnagy / @SabotageAndi ?
  * [x] Print errors as ndjson
  * [x] Replace `bin/gherkin-generate-ast` with a new `bin/gherkin`
  * [x] Implement stream API
  * [ ] Implement a pickles compiler and use it in stream API and `bin/gherkin`

If you want to help, create a branch from the `cli` branch, for example `cli-go` or `cli-perl`, then send a PR against this branch.

If we can get this done and release it in Gherkin 5 I think we'll have basis for some great things to come:
* Much simpler Cucumber implementations (all of the parsing/compiling work happening in Gherkin)
* Adoption of event protocol, which enables shared HTML/Pretty/etc formatters
* Generally more consistent implementations and interop

/cc @tooky @mattwynne @brasmusson @jbpros @sebrose @nodo @charlierudolph @ciaranmcnulty @jenisys @everzet @paoloambrosio - feel free to help here too :-)

Aslak


aslakhellesoy:
I'd like to get this merged to master so it can go into the next release. The reason I want to get it into the release is so we can make easier progress with Cucumber implementations that aren't yet using Gherkin's pickles, as well as implementing the event protocol in Cucumbers. Work on Gherkin-Lint will also be a lot easier if this is released.

The Perl/Objective-C/Go/.NET ports' builds are currently failing because of the unfinished work described above.

In order to make progress I am therefore considering disabling the build of those ports as well as excluding them from releases until they have been updated.

What are your thoughts about this?

SabotageAndi:
@aslakhellesoy What is the timeframe for the next release? We (at SpecFlow) want to finally finish the .NET Core support, but if it is not released, because of this missing feature, it would make us not happy. 
Currently I have no idea how long would it take to implement this feature in the .NET parser. @gasparnagy: any ideas?
Couldn't you just disable the tests for this feature for the .NET parser?

aslakhellesoy:
Hey @SabotageAndi / @gasparnagy

I have updated the TODO list above for .NET. All tests are passing, and the only missing part is the Gherkin compiler and the resulting Pickles structures.

That should be relatively quick to port [from Java](https://github.com/cucumber/gherkin/tree/master/java/src/main/java/gherkin/pickles). The only class that has any logic here is the [Compiler](https://github.com/cucumber/gherkin/blob/master/java/src/main/java/gherkin/pickles/Compiler.java) - the other classes are value objects without behaviour.

This isn't holding up the release, so .NET will be in the next release, whether .NET Core is ready or not.

gasparnagy:
@aslakhellesoy @SabotageAndi I have a nearly finished compiler implementation somewhere. I will check it and try to push it up today. Maybe that helps the dependency problem.

gasparnagy:
@aslakhellesoy Is there a reason for putting the event-related .NET API of Gherkin (files in the `Events` folder) into the main `Gherkin` namespace? 

The problem is that in .NET people typically import the namespaces and reference the classes by short name. Therefore very generic names, like `IEvent` are generally not used/recommended for .NET libraries, as it makes the code unreadable without fully qualified names. Putting all these into the `Gherkin.Events` namespace is somewhat makes this better, but the real solution would be to rename `IEvent` to `IGherkinEvent` (in addition to the specific namespace).

aslakhellesoy:
That's a pure oversight!

aslakhellesoy:
Make all the changes you want :-) There will be other runtime related implementations of the interface in the event protocol (for results) so maybe call it `ICucumberEvent`? We should do the same for Java (without the I)

gasparnagy:
@aslakhellesoy OK. Let's go for `ICucumberEvent` / `CucumberEvent`. I will make the fix for .NET.

PR #246, opened 2016-12-07
Georgian language

dobiedad:
Added support for the Georgian language. 
All words are translated exactly apart from "background" which i translated into context .


aslakhellesoy:
დიდი მადლობა @dobiedad! Will merge this into the next release :-)

aslakhellesoy:
You even updated the language files - good patch!

dobiedad:
The perl file was a pain!

dobiedad:
@brasmusson I manually updated `gherkin-languages.json`and ran the make task but perl kept on failing. The issue was it expected the JSON module to be in a folder called `Cpanel` and for some reason mine was at the root not in that folder.

Issue #243, opened 2016-12-06
.NET: Implement CLI


aslakhellesoy:
This issue is for remaining work of #240 related to .NET:

* [ ] Implement a pickles compiler and use it in stream API and `bin/gherkin`

@gasparnagy / @SabotageAndi - any takers?

aslakhellesoy:
This is done in the new `cucumber/cucumber` monorepo

PR #247, opened 2016-12-08
WIP: .NET: Implement pickles compiler

gasparnagy:
This is the first step for #240 by targeting #243.

The current source drop is a port of the Java compiler version at June 2016. There were a few smaller improvements in the Java complier since then (https://github.com/cucumber/gherkin/commit/26ed643cd5238b277ff297179b8b20d7748186f7), these are not ported yet.

The testing part is also missing.


aslakhellesoy:
Any progress on this @gasparnagy? Would be great to get into one of the next releases.

PR #248, opened 2016-12-18
[Java] Add accessors for languages and dialect keywords.

brasmusson:
To be able to implement the `--i18n` option in Cucumber-JVM, methods to fetch:
* the list of all available languages ([for the `--i18n help` option](https://github.com/cucumber/cucumber-jvm/blob/waiting-for-gherkin-fixes/core/src/main/java/cucumber/runtime/RuntimeOptions.java#L249-L255)), and 
* the list for each keyword for a specific language ([for the `--i18n <language>` option](https://github.com/cucumber/cucumber-jvm/blob/waiting-for-gherkin-fixes/core/src/main/java/cucumber/runtime/RuntimeOptions.java#L265-L286))

is needed.

### DIFF
--- a/java/src/main/java/gherkin/GherkinDialect.java
+++ b/java/src/main/java/gherkin/GherkinDialect.java
@@ -23,11 +23,11 @@ public class GherkinDialect {
 
     public List<String> getStepKeywords() {
         List<String> result = new ArrayList<>();
-        result.addAll(keywords.get("given"));
-        result.addAll(keywords.get("when"));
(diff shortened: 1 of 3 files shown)


aslakhellesoy:
This sounds like an API that would be convenient for all Cucumber implementations. If we add this API to all implementations it would encourage a small step towards a consistent implementation in Cucumbers. I really think a gradual migration like that is the smoothest journey toward something really consistent.

Would you be interested in porting this API to the other implementations?

PR #255, opened 2017-02-20
Russian translation fix

ehpc:
The russian mapping is conceptually wrong in "given" and "when". "Given" is actually more closely relates to "Если", and "When" relates to "Тогда", but not "Если". Because "Если" semantically dictates preconditions, not an event or current conditions. So the closest to "Given When Then" in Russian is "Если Когда Тогда".


aslakhellesoy:
Thanks @ehpc! @everzet agree?

